PR #104, opened 2017-11-01
Parameterize Girder host via entrypoint

craig-willis:
This PR replaces the build-time sed process to replace the Girder API host with a similar replacement process via Docker entrypoint.  Because the index.html is generated and encoded, I've changed the replacement variable name, removing % and changing case.  

I don't have a strategy for leveraging environment variables at container runtime in JS without moving the build process into the entrypoint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,6 @@ COPY bower.json .
 COPY ember-cli-build.js .
 COPY package.json .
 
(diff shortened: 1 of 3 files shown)


Xarthisius:
Could you rebase?

craig-willis:
Duplicate effort

Issue #126, opened 2018-05-08
3.3 Tale Rendering


ThomasThelen:
When visiting the dashboard, the `Tales` section at the top of the page will sometimes not show any tales (even when a filter has not been selected). This corresponds to `3.3 Tale Rendering` in the google document.

To Reproduce:
   1. Clear your cache & cookies
   2. Log into the dashboard
   3. See missing tales

Workaround:
   Apply the `All` filter to see the tales

A screenshot of tales not displaying.
<img width="1674" alt="screen shot 2018-05-08 at 10 16 15 am" src="https://user-images.githubusercontent.com/9289652/39772031-f09092b6-52a8-11e8-9385-f18711027d0b.png">

PR #131, opened 2018-05-11
Fix Missing Tales

ThomasThelen:
My autoformatter changed a few lines, but the ones of interest are in the `searchFilter` function. I wasn't able to reproduce the problem on the new ui, but I was able to under the master branch. I tested the code under master (and it fixed the issue), and I wasn't able to see any difference in the newui branch. In each failure case, `searchView` was empty at the time
`component.set` and `component.paginate` were called.

This resolves issue #126


ThomasThelen:
The second commit should fix issue . #132

PR #145, opened 2018-06-05
Newui

ThomasThelen:
This adds a checkbox on the registration modal dialog that appears when the server is deployed in a development environment.

A flag was added to `environment.js` which is then checked in the handlebars template to determine whether the checkbox should appear.

### DIFF
--- a/app/components/ui/files/registration-modal/component.js
+++ b/app/components/ui/files/registration-modal/component.js
@@ -14,7 +14,7 @@ export default Ember.Component.extend({
     notificationHandler: Ember.inject.service(),
 
     datasources: Ember.A(),
-
+    dev: config.dev,
     num_results: -1,
     error: false,
     errorMessage: '',
@@ -25,6 +25,9 @@ export default Ember.Component.extend({
     name: '',
     repository: '',
     size: '',
+    useDev: '',
+    devUrl: 'https://dev.nceas.ucsb.edu/knb/d1/mn/v2',
+    prodUrl: 'https://cn.dataone.org/cn/v2',
 
     didInsertElement() {
         this._super(...arguments);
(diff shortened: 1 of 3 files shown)


app/components/ui/files/registration-modal/component.js
```diff
@@ -108,9 +110,15 @@ export default Ember.Component.extend({
     },
 
     actions: {
+
+        updateDev(value) {
+            // Called if the `use dev` checkbox is clicked
+            this.set('useDev', value)
```
cyberdante: This is really minor, @ThomasThelen, but I favor ending lines with a semicolon as a rule that I even try enforcing with my editor.

app/components/ui/files/registration-modal/component.js
```diff
@@ -174,10 +188,17 @@ export default Ember.Component.extend({
             Ember.$('#results').removeClass('hidden');
```
cyberdante: I also avoid using jQuery as much as possible. I favor having variables and rendering the handlebars/html-bars sections using these variables, in this case, something like:

```
    {{#if resultsVisible}}
        <div id="results"> 
            ...
        </div>
    {{/if}}
```

ian-taylor: I agree - there are a number of additional handlebars rules we have included in WholeTale using the ember truth helpers too that allow you to do quite complex logic in the templates directly: https://github.com/jmurphyau/ember-truth-helpers

cyberdante: @ThomasThelen, where are we on this? Can I merge and close this PR already?

ThomasThelen: Ready to go!

PR #221, opened 2018-08-01
Publishing modal

### LINKED ISSUE #218: Refactor Modals to Use ember-modal-dialog
We should refactor our modal dialogs to use `ember-modal-dialog`. This will allow us to pass data to the modal (like tale id).

See https://dockyard.com/blog/2015/06/08/a-clean-pattern-for-modal-dialogs for more information on how to do this.

ThomasThelen:
The first change was done to reflect a change in the girder wholetale plugin.
The second change makes the tale ID accessible to the publishing modal dialog. 

For more information on each, see the commit message.

Fixes #218

### DIFF
--- a/app/components/ui/files/publish-modal/component.js
+++ b/app/components/ui/files/publish-modal/component.js
@@ -16,17 +16,16 @@ export default Ember.Component.extend({
     publishingFinish: false,
     // An array that holds a pair, (fileName, fileSize)
     fileList: [],
-    // Holds an array of objects that cannot be excluded from publishing
-    immutaleFile: ['tale.yaml', 'environment.zip'],
+    // Holds an array of objects that the user cannot be exclude from their package
+    nonOptionalFile: ['tale.yaml', 'environment.zip'],
     // A map that connects the repository dropdown to a url
     repositoryMapping: {'Development': 'https://dev.nceas.ucsb.edu/knb/d1/mn/'},
     // The url for the published tale. This is set after publication succeeds
     packageUrl: '',
     // The jwt token that allows the user to interact with DataONE
-    jwt: '',
+    dataoneJWT: '',
 
     setPublishBtnState(state) {
-        console.debug(state)
         this.set('enablePublish', state);
     },
 
@@ -51,30 +50,28 @@ export default Ember.Component.extend({
                     limit: 0,
                     sort: 'lowerName',
                     sortdir: 1
-                }}
+                }};
             self.get('authRequest').send(url) 
             .then(rep => {
                 let fileList2 = []
                 rep.forEach(function(item)
                 {
-                    let path = item.name
-                    fileList2.push({'name': path, 'size':item.size, 'id': item._id})
+                    let path = item.name;
+                    fileList2.push({'name': path, 'size':item.size, 'id': item._id});
                 })
-                self.set('fileList', fileList2)
+                self.set('fileList', fileList2);
             })
         })
-        console.log(self.get('fileList'))
     },
 
     getPath(id) {
-            let url = config.apiUrl + '/item/'+ id + '/rootpath'
+            let url = config.apiUrl + '/item/'+ id + '/rootpath';
             let self = this;
             let path = '';
             let hasRoot = false;
 
             this.get('authRequest').send(url)
                 .then(rep => {
-                    console.log(rep);
                     rep.forEach(function(folder) {
                         let folderName = folder.object['name'];
                         if (folderName == 'Data' | folderName == "Home") {
@@ -85,13 +82,11 @@ export default Ember.Component.extend({
                             path += folderName;
                         }
                     })
-                    console.log(path)
                 });
     },
 
     getFileParent(file) {
         let self = this;
-    
         let promisedParentMeta;
     
         if(file.get('_modelType') === "folder") {
@@ -131,11 +126,11 @@ export default Ember.Component.extend({
     },
 
     joinArray(arr) {
-        let joined = String()
+        let joined = String();
         arr.forEach(function(item) {
-            joined += "'" +item+"'"
+            joined += "'" +item+"'";
         })
-        return joined
+        return joined;
     },
 
     prepareItemIds() {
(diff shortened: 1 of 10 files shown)


app/components/ui/files/publish-modal/component.js
```diff
@@ -6,18 +6,23 @@ import config from '../../../../config/environment';
...
     publishingFinish: false,
     // An array that holds a pair, (fileName, fileSize)
     fileList: [],
     // Holds an array of objects that cannot be excluded from publishing
     immutaleFile: ['tale.yaml', 'environment.zip'],
+    // A map that connects the repository dropdown to a url
     repositoryMapping: {'Development': 'https://dev.nceas.ucsb.edu/knb/d1/mn/'},
+    // The url for the published tale. This is set after publication succeeds
     packageUrl: '',
+    // The jwt token that allows the user to interact with DataONE
```
cyberdante: @ThomasThelen I would have named this token variable differently, just in case we need to add another token for interacting with a different data source or system.

app/components/ui/files/publish-modal/component.js
```diff
@@ -167,26 +172,23 @@ export default Ember.Component.extend({
...
     dataoneLogin() {
         /*
         Responsible for opening the login dialog for the user. Ideally, we could
         tell when the user finishes logging in so that we know when to fetch the token
         */
-       let url = 'https://cn.dataone.org/portal/oauth?action=start&target=http://probable-cattle.nceas.ucsb.edu:4200/redirect'
+       let url = 'https://cn.dataone.org/portal/oauth?action=start&target=http://dashboard.dev.wholetale.org/redirect'
        if (config.dev) {
-        url = 'https://cn-stage-2.test.dataone.org/portal/oauth?action=start&target=http://probable-cattle.nceas.ucsb.edu:4200/redirect'
+        url = 'https://cn-stage-2.test.dataone.org/portal/oauth?action=start&target=http://dashboard.dev.wholetale.org/redirect'
```
cyberdante: Are these `URL`s supposed to be hard-coded?

app/components/ui/files/publish-modal/component.js
```diff
@@ -219,7 +224,6 @@ export default Ember.Component.extend({
     },
 
     openPackage: function() {
-            
         var win = window.open(this.get('packageUrl'), '_blank');
```
cyberdante: Try substituting all your `var` declarations for the correct/updated `let` or `const` on each `.js` file you touch.

app/components/ui/files/publish-modal/component.js
```diff
@@ -231,7 +235,6 @@ export default Ember.Component.extend({
             Called when the `Publish` button is clicked. It controls the flow
             of logging in and communicating with the `createPackage` endpoint.
             */
-
            let self = this;
            if (self.get('publishingFinish') == true) {
```
cyberdante: @ThomasThelen - we should ALWAYS use the less permissive `===` for comparisons. Also, in this case, to check if the expression is _truthy_, you would have just needed to do `if (self.get('publishingFinish')) { ... } `

app/components/ui/files/publish-modal/template.hbs
```diff
@@ -69,16 +70,13 @@
                     <div class="list">
                         {{#each immutaleFile as |file|}}
```
cyberdante: immutable?

app/components/ui/files/publish-modal/template.hbs
```diff
@@ -129,21 +133,21 @@
     {{/ui-dropdown}}
     
     {{#if enablePublish}}
-    <div class="ui positive blue enabled right labeled icon button">
+    <div class="ui positive blue enabled right labeled icon button" style="position:absolute; right:0px;" {{action "publishedClicked"}}>
```
cyberdante: @ThomasThelen instead of adding the inline style, better create a new class or mixin with those styles you need, and add it to the existing ones `ui positive blue enabled right labeled icon button`

app/routes/application.js
```diff
@@ -39,6 +39,22 @@ export default AuthenticateRoute.extend({
...
+    
+        Ember.setProperties(applicationController, {
+          modalDialogName,
+          modalContext,
+          isModalVisible: true
+        });
+      },
+    
+    closeModal() {
+        debugger;
```
cyberdante: @ThomasThelen - showstopper alert here! Cannot merge code with a __debugger;__ statement left behind. With great power comes great responsibility!

ThomasThelen: Ouch good catch

cyberdante: 👍

app/components/ui/files/publish-modal/component.js
```diff
@@ -46,30 +50,28 @@ export default Ember.Component.extend({
...
             let self = this;
             let path = '';
             let hasRoot = false;
 
             this.get('authRequest').send(url)
                 .then(rep => {
-                    console.log(rep);
                     rep.forEach(function(folder) {
                         let folderName = folder.object['name'];
                         if (folderName == 'Data' | folderName == "Home") {
```
cyberdante: this `if` must use `===` and `||` for the _or_ operator, @ThomasThelen 
Sorry I did not catch this earlier

cyberdante: And please, use the _hard_ quotation marks (`' '`) consistently and as much as possible, favor them over the _soft_ ones (`" "`)

app/components/ui/files/publish-modal/component.js
```diff
@@ -160,67 +160,75 @@ export default Ember.Component.extend({
...
-      },
-
       
     dataoneLogin() {
         /*
         Responsible for opening the login dialog for the user. Ideally, we could
         tell when the user finishes logging in so that we know when to fetch the token
         */
-       let url = 'https://cn.dataone.org/portal/oauth?action=start&target=http://probable-cattle.nceas.ucsb.edu:4200/redirect'
+       let url = 'https://cn.dataone.org/portal/oauth?action=start&target=http://dashboard.dev.wholetale.org/redirect';
```
cyberdante: These URLs should come from the `environment/config` file and be set differently for __development__ and __production__

PR #222, opened 2018-08-04
Touch up License Section

ThomasThelen:
The changes in this PR are related to the license section in the publishing modal dialog.

## Change 1

I added the popups to the info icons that are next to the license names and to the one in the main title. This is done in `create_tooltips` which gets called after template rendering occurs.

## Change 2

The license was previously getting reported as CC0 no matter what the user picked. I added a method,
`getSelectedLicense` that gets the id of the selected radio button. I added an `id` field to each radio element that ranges from 0-2 (each number represents a different license type).

## Change 3

I added the images next to the license choices that are shown in the mockup.


app/components/ui/files/publish-modal/component.js
```diff
@@ -173,8 +213,8 @@ export default Ember.Component.extend({
         Responsible for opening the login dialog for the user. Ideally, we could
         tell when the user finishes logging in so that we know when to fetch the token
         */
-       let url = 'https://cn.dataone.org/portal/oauth?action=start&target='+ENV.authRedirect;
-        let newwindow=window.open(url,'auth','height=400,width=450');
+       let url = 'https://cn-stage-2.test.dataone.org/portal/oauth?action=start&target='+config.authRedirect;
```
amoeba: Does the Dashboard have any central configuration store that this URL could be factored out into?

app/components/ui/files/publish-modal/template.hbs
```diff
@@ -97,28 +99,37 @@
...
-    <div class="ui form"></div>
+    <div class="ui form">
         <div class="grouped fields">
             <div class="field">
-                <div class="ui radio checkbox">
-                    <input type="radio" name="CC0" checked="checked">
-                    <label>Creative Commons Public Domain CCO <i class="info circle blue icon" ></i> </label> 
+                <div class="ui radio checkbox cc0" id='0'>
+                    <input type="radio" name="license-radio" checked="checked">
+                    <label><img class="ui small image left floated" src="https://d15omoko64skxi.cloudfront.net/wp-content/uploads/2014/01/cc-zero.png">
```
amoeba: I don't know how others feel, but maybe referencing images on external servers is likely to break at some point or annoy the respective image hosts. What do you think about including these images as formal assets in this repo?

app/components/ui/files/publish-modal/template.hbs
```diff
@@ -97,28 +99,37 @@
...
 
 <div class="content">
-    <div class="ui form"></div>
+    <div class="ui form">
         <div class="grouped fields">
             <div class="field">
-                <div class="ui radio checkbox">
-                    <input type="radio" name="CC0" checked="checked">
-                    <label>Creative Commons Public Domain CCO <i class="info circle blue icon" ></i> </label> 
+                <div class="ui radio checkbox cc0" id='0'>
```
amoeba: Is this an Ember-y way to do this? Does Ember tend to grab values off the DOM when they're needed or does it use more of a two-way-data-bindy pattern? This is where my knowledge of Ember kinda breaks down but I wonder if this might instead be written more like:

```js
{{#each availableLicenses as |license|}}
  ...
  <input ... {{action "updateLicense" license}}
  ...
{{/each}}
```

So that the license is already set when it's needed.

app/components/ui/files/publish-modal/component.js
ThomasThelen: I can stick this in `environment.js`

amoeba: 👍

ThomasThelen: Addressed in 3e8b9afb676cac4146bc21e1330f77616c15db60

ThomasThelen:
This is moving to a new PR

PR #223, opened 2018-08-05
Environment Section Touch Up

ThomasThelen:
I branched this branch off of `dataone_publishing` which at the time of this PR is still a PR (so you'll see extra commits). The commit that needs to be reviewed is afd6667

## Change 1

I added the license file to the environment files section

## Change 2

I added tooltips for each environment file, briefly describing what that are for. To get the selection right, I added the file name to the end of the div class.

### Change 3

Styling changes to the environment section. I changed the size of the folder and the header text. In addition, I made the header text the name of the tale, which I retrieve from the `tale` endpoint.

Here's a screenshot of what it looks like after these changes.

screen shot 2018-08-04 at 5 12 06 pm

### DIFF
--- a/app/components/ui/files/publish-modal/component.js
+++ b/app/components/ui/files/publish-modal/component.js
@@ -213,8 +213,8 @@ export default Ember.Component.extend({
         Responsible for opening the login dialog for the user. Ideally, we could
         tell when the user finishes logging in so that we know when to fetch the token
         */
-       let url = 'http://cn-stage-2.test.dataone.org/portal/oauth?action=start&target='+config.authRedirect;
-       let newwindow=window.open(url,'auth','height=400,width=450');
+       let url = 'https://cn-stage-2.test.dataone.org/portal/oauth?action=start&target='+config.authRedirect;
+       let newwindow = window.open(url,'auth','height=400,width=450');
     },
 
     loggedIntoDataONE() {
--- a/app/components/ui/files/publish-modal/component.js
+++ b/app/components/ui/files/publish-modal/component.js
@@ -17,16 +17,39 @@ export default Ember.Component.extend({
     // An array that holds a pair, (fileName, fileSize)
     fileList: [],
     // Holds an array of objects that the user cannot be exclude from their package
-    nonOptionalFile: ['tale.yaml', 'environment.zip'],
+    nonOptionalFile: ['tale.yaml', 'environment.tar', 'license.txt'],
     // A map that connects the repository dropdown to a url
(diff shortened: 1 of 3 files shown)


app/components/ui/files/publish-modal/component.js
```diff
@@ -17,18 +17,41 @@ export default Ember.Component.extend({
     // An array that holds a pair, (fileName, fileSize)
     fileList: [],
     // Holds an array of objects that the user cannot be exclude from their package
-    nonOptionalFile: ['tale.yaml', 'environment.zip'],
+    nonOptionalFile: ['tale.yaml', 'environment.tar', 'license.txt'],
```
amoeba: I assume we'll be archiving a tar.gz and not just a tar. Also, `license.txt` is now LICENSE.

app/components/ui/files/publish-modal/component.js
```diff
@@ -17,18 +17,41 @@ export default Ember.Component.extend({
...
+        // Create the tooltips after the template has been rendered
+        this.create_tooltips();
+    },
 
     setPublishBtnState(state) {
         this.set('enablePublish', state);
     },
 
+    setTaleName() {
+        let url = config.apiUrl + '/tale/'+ this.get('modalContext')
```
amoeba: I thought I commented on this yesterday but can't seem to find it...

Is there a more Ember-y way to do this? I see two things that struck me as odd: (1) the URL for the request is constructed manually and this results in the user's browser sending an extra request for the Tale information when it already has it, in theory.

For (1), I'd expect something more concise like:

```js
self.set('taleName', new Tale(this.get('modalContext').get('title'))
```

For (2), does Ember let you give the modal the Tale information when it's built or can it peek outside itself, effectively sharing the information that was retrieved outside the modal's lifecycle?

amoeba: Oh, I did make this comment. Found it! https://github.com/whole-tale/dashboard/pull/222 Nvmd.

ThomasThelen: I'll rebase/merge this branch with the parent with my next commit if that helps

app/styles/modals.scss
```diff
@@ -126,4 +126,12 @@
 .publisher .ui.radio.checkbox {
     left:3%;
     font-size:medium;
-}
\ No newline at end of file
+}
+
+.publisher .environment.folder.icon {
+    font-size: 25;
```
amoeba: missing a unit

ThomasThelen:
This is moving to a new PR

Issue #171, opened 2018-06-28
Show Files That Can't Be (un)selected in the  `Environment` Section in the Publishing Modal


ThomasThelen:
We should show the user which additional files are getting added into the package such as the tale.yaml and repository.tar files..

ThomasThelen:
Finished in PR #229

PR #227, opened 2018-08-06
Misc Updates to PublishingModal

ThomasThelen:
There are a couple misc changes in here, mostly css changes

# Change 1

Misc styling things like the font color and alignment changes.

# Change 2

I added the `science_metadata.xml` file to the Environment section. Should complete #171 

# Change 3

The dialog now listens on the event stream for any publishing notifications.

Relevant commits are

3adab8dbd72defd40e7a8bcb449ef9ae6371b62a
afd66674701d108142d9553054031557840e25bf
0e8cf4025cc4bdffeeffd7e2d00b62bd713292bb
3aed0bc6fd2333a6852eb4420b7b12b7313a30fe
b3a375d168cc8f0bb3ea43e8a95a37c1d775335b

### DIFF
--- a/app/components/ui/files/publish-modal/component.js
+++ b/app/components/ui/files/publish-modal/component.js
@@ -213,8 +213,8 @@ export default Ember.Component.extend({
         Responsible for opening the login dialog for the user. Ideally, we could
         tell when the user finishes logging in so that we know when to fetch the token
         */
-       let url = 'http://cn-stage-2.test.dataone.org/portal/oauth?action=start&target='+config.authRedirect;
-       let newwindow=window.open(url,'auth','height=400,width=450');
+       let url = 'https://cn-stage-2.test.dataone.org/portal/oauth?action=start&target='+config.authRedirect;
+       let newwindow = window.open(url,'auth','height=400,width=450');
     },
 
     loggedIntoDataONE() {
--- a/app/components/ui/files/publish-modal/component.js
+++ b/app/components/ui/files/publish-modal/component.js
@@ -17,16 +17,39 @@ export default Ember.Component.extend({
     // An array that holds a pair, (fileName, fileSize)
     fileList: [],
     // Holds an array of objects that the user cannot be exclude from their package
-    nonOptionalFile: ['tale.yaml', 'environment.zip'],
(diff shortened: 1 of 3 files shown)


ThomasThelen:
screen shot 2018-08-06 at 9 11 03 am
screen shot 2018-08-06 at 9 10 59 am


How it looks after this PR

app/components/ui/files/publish-modal/component.js
```diff
@@ -17,18 +19,44 @@ export default Ember.Component.extend({
...
+        // Create the tooltips after the template has been rendered
+        this.create_tooltips();
+    },
 
     setPublishBtnState(state) {
         this.set('enablePublish', state);
     },
 
+    setTaleName() {
+        let url = config.apiUrl + '/tale/'+ this.get('modalContext')
```
amoeba: Feel free to not trouble yourself over this as my Ember experience is very much in progress. I remember asking others on Slack for the most Emberish way to do this so feel free to disregard my ramblings if this is just the way others are doing this type of stuff.

Does Ember give us a handy-dandy way to do this without (1) the extra call to `/tale` and without having to construct the URL from scratch? Re: (1), I wonder if the underlying view can share the already-fetched Tale instance with the modal to save the extra call. Re: (2) I'm used to applications using reflection of some sort to dynamically determine routes so the code is resilient to changes in URL structure and, in this case, delegating the responsibility of determining the URL and doing the querying to other parts of the application. It'd result in code like:

```js
setTaleName() {
  var tale = new Tale(this.get('modalContext')) # Handles the request
  self.set(tale.get('title'))
}
```

amoeba:
This is all looking great. Is the icon that was chosen for the  Usage Rights section of the modal intentional?

Since #222 and #223 contain so many related/shared commits across similar paths, what do you think about either combining these into a single PR or something? It's totally fine if, in the future, if you want to keep pushing new commits to an existing PR if the commits are tightly related.

ThomasThelen:
That icon from semanticui is broken (I tried it in a new ember application) so I'm going to switch them over to font awesome (we're moving away from semanticui)

ThomasThelen:
Moving to a new PR

PR #260, opened 2018-08-30
Added dataOneHOST and dashboardHOST variables and replacements

craig-willis:
See also https://github.com/whole-tale/terraform_deployment/pull/36
* Added ember dataOneHost env
* Added replacement for dashboardHOST in environment.js and docker-entrypoint.sh


ThomasThelen:
We can actually now do away with the `ENV.authRedirect` variable

craig-willis:
@ThomasThelen Removed the authRedirect variable.

PR #262, opened 2018-09-05
Fix file picker jumpiness by fixing the first two columns

### LINKED ISSUE #261: File picker columns and content move around unpredictably
The new file picker works pretty well for the most part but I noticed that the columns resize around seemingly arbitrarily as I expand and collapse folders. It'd be a lot nicer for the user if things didn't jump around.

amoeba:
This is such a small change maybe we don't need another set of eyes but just in case I'll leave this open until tomorrow afternoon.

Old behavior:

kapture 2018-09-05 at 15 32 30

New behavior:

kapture 2018-09-05 at 15 33 29

Closes #261


amoeba:
Thanks @ThomasThelen !

PR #264, opened 2018-09-11
Fix #255. Distinguish when no instance is selected

### LINKED ISSUE #255: Disable Publish & Edit Buttons in Run Paddleboard on Empty Tale
screen shot 2018-08-27 at 6 01 06 pm

We should probably grey out/hide the two buttons when the `/run` page is visited if a tale hasn't been selected. 

This can be done by checking the model for the tale ID. If it doesn't exist, disable the buttons.

cyberdante:
Added changes to distinguish the route `run.index` (no instance selected) from `run.view`


ThomasThelen:
There's a debugger statement at
https://github.com/whole-tale/dashboard/pull/264/files#diff-97e5c5940e407cb2d5486714e28ce631R15

PR #273, opened 2018-09-19
Run tale from view

cyberdante:
Please, @ThomasThelen make sure to run this branch and check that the issues tagged as fixed by it are indeed solved. And as usual, thanks, mate!


ThomasThelen:
@cyberdante In `app/controllers/compose.js` lines 40&41, do we need to set these after they were changed into computed properties in lines 24&27?

PR #275, opened 2018-09-25
 Allow to specify OAuth provider via env var

Xarthisius:
This PR allows to modify the OAuth provider we're using in the dashboard during runtime.


ThomasThelen:
LGTM-thanks for trimming `environment.js`

PR #277, opened 2018-09-27
Bring back dev config environment

Xarthisius:
As noted by @amoeba, it's useful to have sane defaults for local development.


amoeba:
Thanks for doing this so quickly. Tested locally and it's working nicely.

Issue #285, opened 2018-11-08
Refactor view/edit metadata based on new mockups


craig-willis:
Based on the latest UI revisions suggested by Kristina, we will be moving the current view/edit tale metadata functionality from the browse page to the Run page for a tale instance.

Based on the [10/31 planning meeting](https://wholetale.readthedocs.io/development/meetings/2018-10-31.html#edit-view-tale-metadata) this includes:
* Adding "Metadata" tab to Run page
* Refactor view/edit page per mockup
* Out of scope for this issue: publications, browse data, license, derived from, versions, dataset citation information (requires backend changes)
* If user has edit access, they should be able to modify:
  * Title, authors, categories, description, image, change environment
* The following fields are read-only: 
  * Datasets used is derived from current state of data
  * DOI-level citation or other human-readable format for top-level registered item

See also:
* Mockups  for redesign ["Metadata" panel](http://www.crc.nd.edu/~kfurse/wholetale-css-mock-up-sui_f2018/run-meta.html)
* [Invision flow](https://projects.invisionapp.com/freehand/document/nPpPcHSRg) Fall 2018 revisions section

ThomasThelen:
This might be connected to 
https://github.com/whole-tale/dashboard/issues/256

ThomasThelen:
This was discussed on a dev call with @craig-willis, @Xarthisius, and @bodom0015 on 14 November 2018.

I believe the only part that was left undecided was whether we should still follow @cyberdante 's recommendation in issue #256 about "componentizing" the controller instead of just embedding it.

I should note that the traditional Tale view is going away, and the code will only be used in one place-correct me if I'm wrong.

bodom0015:
I have started copying the code out of the existing controller into a separate space, which could easily be wrapped in a component if that might be preferable.

While playing around with EmberJS/Semantic UI forms, I created a dropdown to list out the environments. Unfortunately, it seems that we do not currently expose `imageId` as a modifiable field on the Tale model, so my changes do not actually persist when saving to Girdert: https://github.com/whole-tale/girder_wholetale/blob/306e42d30e02238e05152cd70c784b9d33975e65/server/models/tale.py#L34-L43

cyberdante:
@ThomasThelen I don't know how this was tackled but have a few minutes to address my recommendation of making this code a component:
[...]

PR #297, opened 2018-11-26
Migrate "View/Edit Tale Metadata" functionality to the Run view

bodom0015:
Changes:
* Added Interact / Metadata tabs to the "Run" view
    * Interact Tab: the previous "Run" view can be found here - iframe or link to the wholetale frontend/instance
    * Metadata Tab: a new view based heavily on the previous "View/Edit Tale" view can be found here

Test Case

1. Checkout this branch locally, pointed at `girder.dev.wholetale.org`: `git fetch --all && git checkout refactor-view-edit-tale-metadata`
2. Start Ember: `ember s --environment=production`
3. Login to the dev instance that you just started
4. Launch a Tale that you own - create one if needed
5. Select the Tale Instance that you just launched and choose "Run" at the top
    * You should see the "Interact" and "Metadata" tabs
    * The Interact tab should be selected by default, which should show the familiar Tale view (e.g. iframe or link)
6. Click the "Metadata" tab
    * Tab content should change to offer "View/Edit Tale Metadata" functionality
    * If you do not own or cannot edit the Tale, the form should be read-only
    * Even if you do own the Tale, `Date Created`, `Last Updated`, and `Authors` should still be read-only
    * Valid environments should be listed in the form dropdown for `Environment`
7. Change the name of the Tale and click "Save"
    * NOTE: The Environment field will not currently save - see https://github.com/whole-tale/dashboard/issues/285#issuecomment-440426589
8. Refresh the page
    * Your changes should be preserved between page refreshes, confirming that they are being properly persisted to the server


bodom0015:
Refactored the tab structure into components, squashed some commits, and cleaned things up a bit.

Hopefully this makes for a quicker review process

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,91 @@
...
+  }).readOnly(),
+  cannotEditTale: computed.not('canEditTale').readOnly(),
+  
+  actions: {
+    // TODO: Status messages are still missing
+    updateTale() {
+      let tale = this.get("model").get("tale");
+      let component = this;
+      //component.set("tale_creating", true);
+      console.log("Sending updated tale:", tale);
```
ThomasThelen: Small nitpicking; I believe we have an eslint `no-console` rule from the .eslintrc file that we've been trying to follow. This may have come from the controller code, but we generally try to update old code to follow the current linter settings.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,91 @@
...
+      let onFail = function (e) {
+      // deal with the failure here
+      //component.set("tale_creating", false);
+      //component.set("tale_not_created", true);
+      //component.set("error_msg", e.message);
+      
+      later((function () {
+        //component.set("tale_not_created", false);
+      }), 10000);
+      console.log("Error saving tale (" + tale['id'] + "):", tale)
```
ThomasThelen: This may not be directly related to this PR, but I'll comment here for reference. We might want to notify the user that their Tale failed to save and to try again. If it's not new code we can probably put an issue in for it, if it's needed. @craig-willis what do you think?

app/components/ui/tale-tabs-selector/component.js
```diff
@@ -0,0 +1,30 @@
+import Ember from 'ember';
+
+
+export default Ember.Component.extend({
```
ThomasThelen: We should be using import Component from '@ember/component';
and using it as export default Component.extend({
See this file as an example.

This satisfies the linter's new-module-imports rule. There are other places in the code that are still using Ember.Component that we've been replacing as we run across them.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,91 @@
...
+    updateTale() {
+      let tale = this.get("model").get("tale");
+      let component = this;
+      //component.set("tale_creating", true);
+      console.log("Sending updated tale:", tale);
+      console.log("New image ID:", tale.imageId);
+      
+      let onSuccess = (item) => {
+      //component.set("tale_creating", false);
+      //component.set("tale_created", true);
```
ThomasThelen: We can probably get rid of these comments since tale_created, tale_creating, tale_not_created, and error_msg aren't used here or in the template.

app/components/ui/tale-tab-metadata/template.hbs
```diff
@@ -0,0 +1,91 @@
...
+        {{ui/markdown-editor markdownValue=model.tale.description}}
+      </div>
+  </div>
+  
+  <div class="inline fields ui grid">
+    <label class="two wide column right aligned">Illustration</label>
+    <div class="ui action field thirteen wide column">
+        {{input type="text" id="tale-icon" placeholder="Tale Icon" value=model.tale.illustration readonly=cannotEditTale }}
+        <div class="ui buttons">
+            <style type="text/css">
```
ThomasThelen: I know we do inline styling 226 other places but I believe we're trying to consolidate it to the files in the styles directory. Small things like display, font-size, etc should be fine inline. @cyberdante correct me if I'm wrong though.

app/components/ui/tale-tab-metadata/component.js
bodom0015: Does EmberJS offer a lint-friendly way to write to the console?

With AngularJS there was `$log` service that provided a bit more robust logging support without directly relying on `console.log`

app/components/ui/tale-tabs-selector/component.js
bodom0015: > See this file as an example.

Was there supposed to be a link here? In any case, I am happy to loop thru and update the few new components with the preferred pattern

app/components/ui/tale-tab-metadata/component.js
bodom0015: Understood! I wasn't sure if the long-term plan was to move them over or to just get rid of them, but I can at least remove them for now

app/components/ui/tale-tab-metadata/template.hbs
bodom0015: This was lazily copied over from the original here: https://github.com/whole-tale/dashboard/blob/master/app/templates/forms/tale-edit.hbs#L39-L44

In my experience with other component-based architectures has been to put the per-component styling in the same directory as the component itself... If we would rather abstract it to a `.css` file elsewhere, that can be arranged as well

app/components/ui/tale-tab-metadata/component.js
bodom0015: I can easily change the `console.log` to a classic `alert()` for the time being. There is apparently a warning modal that we may be able to reuse if we would rather have fancy styling on the error message

app/components/ui/tale-tab-metadata/component.js
ThomasThelen: I know EmberJS provides the following [documentation](https://guides.emberjs.com/release/configuring-ember/debugging/) on debugging Ember apps but it looks like a fair amount of overhead.


I usually use console.log because it's quick and then remove it after I'm done.

app/components/ui/tale-tab-metadata/template.hbs
ThomasThelen: I'm definitely guilty of not following this pattern, but at the bottom of our README we mention `ember-cli-component-pod` which we should probably be using. Since it's not new code I'm fine with leaving it as-is.

app/components/ui/tale-tab-metadata/component.js
bodom0015: Interesting.. this documentation seems to be more concerned with tuning the existing logging done by Ember in the background. I guess I'm not sure why using `console.log` would be discouraged is there is no framework-specific substitute.

Although I definitely understand the desire to keep these at a minimum and to remove any leftover debug messages before merging PRs, there are certainly times (such as in the `onFail` function below) where we may want to log error messages from the server. Is there an example of the accepted pattern somewhere that I could take a look at?

ThomasThelen: It seems rather esoteric to me. You can find a handful of blog posts that argue not to use them, and the [docs](https://eslint.org/docs/rules/no-console) don't give a reason other than `Such messages are considered to be for debugging purposes and therefore not suitable to ship to the client`.

I don't know of any better way to handle a failure from the server other than an error message popup. In more complicated parts of the codebase I could see leaving commented out logging statements in. @cyberdante, what's your opinion on `console.log`?

app/components/ui/tale-tab-metadata/component.js
ThomasThelen: I wasn't aware of `Ember.onerror`; it looks pretty handy. We could put a common error dialog into a service and inject it into the various components, but that sounds like a separate task.

app/components/ui/tale-tabs-selector/component.js
ThomasThelen: Oops it should have been [here](https://github.com/whole-tale/dashboard/blob/9b948c667ed4c7fdbe975c081db4191cc82802c2/app/components/ui/tale-instances/component.js#L9)

app/components/ui/tale-tab-metadata/template.hbs
cyberdante: @bodom0015 @ThomasThelen: Ember did not provide a specific way of combining in a per-component basis folder every `<component_name>.hbs`, `<component_name>.(s)css`, and `<component_name>.js`, which I also favor. Its original folder structure follows the one by Rails, which was an inspiration for Ember.
Nonetheless, [pods](https://www.npmjs.com/package/ember-cli-component-pod) attempted to solve this situation shortly thereafter. I have personally used this addon [ember-component-css](https://github.com/ebryn/ember-component-css) for a similar contained structure to that previously mentioned, but instead, having:
```
    - app
        - components
            - <component_name>
                component.js
                template.hbs
                styles.scss
```
It is not ideal, but it provides an improvement. It needs some tweaks to the `.scss` files, mainly wrapping the whole file with a certain class (I usually mimic the component's name in `kebab case` and add a `classNames: ['<component_name>']` to the `component.js` file).

Finally, the community was divided with regard to using PODS. Mainly because this structure cannot be ported to other projects seamlessly, but also because it does not work well with addons. Hence, the appearance of _module unification_ and the adoption of _BEM_ for the styles. Sadly, this project had already started when I joined it, and the effort we can devote is driven by the budget that Notre Dame has assigned for my team to work on it. At this point, modifying the whole UI would require a major iteration for which we don't have enough funding, in my opinion.

app/components/ui/tale-tab-files/component.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+
+
+export default Component.extend({
+  model: null,
+  init() {
+    this._super(...arguments);
+  },
+  
+  actions: {
```
cyberdante: If no actions performed by this component, I'd rather not include the `actions` hook. This is not a mistake, just more lines for the compiler to process, which is already slow as it is.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+  cannotEditTale: computed.not('canEditTale').readOnly(),
+  
+  actions: {
+    updateTale() {
+      let tale = this.get("model").get("tale");
+      let onSuccess = (item) => {
+        //console.log("Successfully saved tale (" + tale['id'] + "):", tale);
+      };
+      
+      let onFail = (e) => {
```
cyberdante: The linter will likely complain here about variable `e` which is declared but never used.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+  
+  canEditTale: computed('model.tale._accessLevel', function () {
+    return this.get('model') && this.get('model').get('tale') && this.get('model').get('tale').get('_accessLevel') >= taleStatus.WRITE;
+  }).readOnly(),
+  cannotEditTale: computed.not('canEditTale').readOnly(),
+  
+  actions: {
+    updateTale() {
+      let tale = this.get("model").get("tale");
+      let onSuccess = (item) => {
```
cyberdante: The linter will likely complain here about variable `item` which is declared but never used.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+      
+      let onFail = (e) => {
+        alert("Error saving tale (" + tale['id'] + "):", tale);
+      };
+      
+      tale.save().then(onSuccess).catch(onFail);
+    },
+      
+    setTaleEnvironment: function(selected) {
+      let tale = this.get('model').get("tale");
```
cyberdante: Let's try using single quotes as much as possible, and not mixing them up with double quotes.

app/components/ui/tale-tab-metadata/template.hbs
```diff
@@ -0,0 +1,91 @@
...
+    <label class="two wide column right aligned">Environment</label>
+    <div class="field thirteen wide column">
+      <div class='ui icon selection dropdown {{if (eq environments.length 0) "loading"}}' style='width:100%'>
+        <input name='TaleImageId' type='hidden' onchange={{action "setTaleEnvironment" value="target.value"}}>
+        <i class='dropdown icon'></i>
+        <div class='default text'> Choose an environment...</div>
+        <div class='menu'>
+          {{#each environments as |env|}}
+            <div class='item' data-value={{env._id}}>
+              <img class='ui image' src={{env.icon}} style='height:28px;width:28px;'> {{env.name}}
```
cyberdante: Favor classes usage over inline style

app/components/ui/tale-tab-metadata/template.hbs
```diff
@@ -0,0 +1,91 @@
...
+          {{/each}}
+        </div>
+      </div>
+    </div>
+  </div>
+  
+  <div class="inline fields ui grid">
+    <label class="two wide column right aligned">Date Created</label>
+    <div class="field thirteen wide column">
+        {{input value=model.tale.created readonly="true"}}
```
cyberdante: @bodom0015 according to the mockups, editable fields should be `input` and `select` elements, but readonly ones should just be labels, not to confuse the user as to suggest that they can be later be enabled for edition under a condition that might change.
screen shot 2018-12-06 at 10 16 26 am

app/components/ui/tale-tab-metadata/template.hbs
```diff
@@ -0,0 +1,91 @@
...
+    </div>
+  </div>
+  
+  <div class="inline fields ui grid">
+    <label class="two wide column right aligned">Last Updated</label>
+    <div class="field thirteen wide column">
+      {{input value=model.tale.updated readonly="true"}}
```
cyberdante: Ditto for `readonly` fields

app/components/ui/tale-tabs-selector/template.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="ui">
+  <div class="ui secondary pointing menu" style="display:inline-flex;">
+      <a class="item {{if activeTabInteract "active"}}" data-tab="tab-interact" onclick={{action "activateInteract"}}>Interact</a>
+      <!--<a class="item {{if activeTabFiles "active"}}" data-tab="tab-files" onclick={{action "activateFiles"}}>Files</a>-->
```
cyberdante: Comment in `.hbs` files should be made with {{!-- --}}, otherwise there are instances where things break when they are compiled to `.js` files by Ember

app/components/ui/tale-tabs-selector/template.hbs
```diff
@@ -0,0 +1,22 @@
...
+      {{#if activeTabInteract}}
+        {{ui/tale-tab-interact model=model}}
+      {{/if}}
+  </div>
+  <div class="ui tab {{if activeTabFiles "active"}}" data-tab="tab-files">
+      {{#if activeTabFiles}}
+        {{ui/tale-tab-files model=model}}
+      {{/if}}
+  </div>
+  <div class="ui tab {{if activeTabMetadata "active"}}" data-tab="tab-metadata" style="overflow-x:auto;max-height:50vh;">
```
cyberdante: screen shot 2018-12-06 at 8 37 36 am

cyberdante:
By the way, @bodom0015 I had to modify the [`config/environment.js`](https://github.com/whole-tale/dashboard/blob/2be233caac7698e136df3af4f1b1f018dcb566f5/config/environment.js#L84) file to be able to test these changes per your instructions. 
Also, although they are hand-to-hand as of now, we would probably want to merge these changes into the [`feature/tale-workspace-support`](https://github.com/whole-tale/dashboard/tree/feature/tale-workspace-support) branch

app/components/ui/tale-tab-files/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+        <div class="ui collapsable grid folder-grid">
+          {{ui/files/folder-navigator currentNavCommand=currentNavCommand openUploadDialog="openUploadDialog" openCreateFolderModal="openCreateFolderModal" onRegisterDataset="openRegisterModal" navClicked="navClicked"}}
+        </div>
+    </div>
+    <div class="eleven wide column">
+        <div class="ui collapsable grid" style="margin-left: 10px">
+            <div class="sixteen wide right aligned column">
+              {{ui/files/directory-browser currentNav=currentNav folderList=fileData.folderContents fileList=fileData.itemContents onItemClicked=(action 'itemClicked')}}
+            </div>
+            <div class="sixteen wide column overlay">
```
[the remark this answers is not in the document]
bodom0015: Forgive my ignorance, but wouldn't the resulting grid look really strange if the number of columns varied in one place, but not in another? It seemed like the upper columns, whether the condition is true or false, is supposed to match the boundaries of the grid below, which is always shown.

Perhaps I am misunderstanding the initial/intended design here?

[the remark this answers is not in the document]
bodom0015: What do you mean here by "the condition"? This line appears to be a component... which part of the line is a conditional?

app/components/ui/tale-tab-files/template.hbs
```diff
@@ -0,0 +1,32 @@
+<div class="main-row">
+    <div class="ui grid">
+        <div class="five wide column">
+            {{#if currentFolderId}}
```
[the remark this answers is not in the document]
bodom0015: It seemed like the columns in the upper grid here, whether currentFolderId is defined or not, was supposed to match the boundaries of the grid below, which is (I think) always shown.

Perhaps I am misunderstanding the initial/intended design here?

app/components/ui/tale-tab-files/template.hbs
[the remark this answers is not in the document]
bodom0015: Ah, I see now where you mean... GitHub didn't render the line below this one in the conversation tab. Can take care of this

app/components/ui/tale-tab-interact/component.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+
+
+export default Component.extend({
+  model: null,
+  init() {
+    this._super(...arguments);
+  },
+  
+  actions: {
```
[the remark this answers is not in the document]
bodom0015: This was a stub, as we will obviously be adding functionality to this page in the near future.

app/components/ui/tale-tab-files/component.js
bodom0015: This was a stub, as we will obviously be adding functionality to this page in the near future.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+  selectDefaultImageId() {
+    // Select the current imageId by default
+    let selectedImageId = this.get("model").get("tale").get("imageId")
+    $('.ui.dropdown').dropdown('set selected', selectedImageId);
+  },
+  
+  canEditTale: computed('model.tale._accessLevel', function () {
+    return this.get('model') && this.get('model').get('tale') && this.get('model').get('tale').get('_accessLevel') >= taleStatus.WRITE;
+  }).readOnly(),
+  cannotEditTale: computed.not('canEditTale').readOnly(),
```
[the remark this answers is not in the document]
bodom0015: We can, but why would we perform an extra import when we don't need to? Having both imports feels like it would make for a slower build, since we potentially need to bundle two copies of the module. Even if webpack is smart enough to realize there is a duplicate, it would probably take longer to figure this out and remove the duplicate import during the compilation.

app/components/ui/tale-tab-metadata/template.hbs
bodom0015: The current mockups, including the screenshot you posted here, do not include **any** information about which fields are editable and which are not. Unless you are seeing a different picture than I am?

bodom0015:
Ah.. would it have worked without the `--environment=production`? I think it should have said  to use `--environment=development` instead, as that seems to point to the existing API.

app/components/ui/tale-tabs-selector/template.hbs
bodom0015: Any tips for how to get the area with the scroll bar to resize correctly with screen size? `height: 100 %` doesn't have the desired effect, as the scroll bar doesn't show up if `max-height` isn't set.

app/components/ui/tale-tab-metadata/component.js
bodom0015: Now consuming the `e` here.. NOTE: this error handler will likely be replaced in the near future with a more generic method for reusing @ThomasThelen's error modal that was introduced in #287.

app/components/ui/tale-tab-metadata/component.js
bodom0015: Removed the success handler, since it was a noop anyway for now

app/components/ui/tale-tab-interact/template.hbs
```diff
@@ -0,0 +1,21 @@
+{{#if (not-eq model.status 0)}}
+    {{#unless model.iframe}}
+        <div class="placeholder message">
+            <p>The compute environment for this Tale must run in a separate browser window.<br/>
+              <a class="ui primary button huge" href="{{model.url}}" target="_blank">Go to new window</a></p>
+
+        </div>
+    {{else}}
+      <div class="rounded-bottom" style="padding: 0; margin: 0; overflow-y: hidden;">
```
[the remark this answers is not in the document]
bodom0015: I have moved these styles to `app/styles/tale-tabs.scss` for the time being.. it should be trivial to move these CSS rules elsewhere later, if a better spot is chosen.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+export default Component.extend({
+  apiHost: config.apiHost,
+  environments: [],
+  model: null,
+  init() {
+    this._super(...arguments);
+    
+    // Fetch images for user to select an environment
+    let component = this;
+    $.getJSON(this.get('apiHost') + '/api/v1/image/').then(function(images) {
```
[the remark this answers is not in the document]
bodom0015: With services: I can perhaps create a service for this single GET request, but it seems like overkill... Doesn't Ember Data provide a way for me to call query a RESTFUL resource? In this case, we aren't updating or deleting, simply listing all entries. 

With route models: Due to how the application is currently architected, we don't necessarily have access to the full `model` provided by `RSVP.hash` in the routes...

For example, assuming that (`app/templates/run/view.hbs`)[https://github.com/whole-tale/dashboard/blob/master/app/templates/run/view.hbs] is still in use, the `left-panel` child's `model` seem to be restricted to the parent's `model.selected`, while the `right-panel` child's `model` is restricted to the parent's `model.instances`. Adding a new value to the `RSVP.hash` in this case, does not allow the `left-panel` to access it. I can change it so that the `left-panel` has access to the full model instead, but that seems like a change we should discuss in case I'm unaware of the repercussions thereof.

app/components/ui/tale-tab-interact/template.hbs
```diff
@@ -0,0 +1,21 @@
+{{#if (not-eq model.status 0)}}
+    {{#unless model.iframe}}
```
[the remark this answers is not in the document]
bodom0015: I have changed the unless/else to an if/else

app/components/ui/tale-tabs-selector/template.hbs
bodom0015: Replaced `<!-- -->` with `{{!-- --}}` here.. I think that's my only comment in a template for now...

I went ahead and stubbed the component here for the new Data tab in hopes that it would prevent some future merge conflict.

app/components/ui/tale-tab-metadata/component.js
```diff
@@ -0,0 +1,68 @@
...
+  
+  didRender() {
+    // Enable the environment dropdown functionality
+    $('.ui.dropdown').dropdown();
+    this.selectDefaultImageId();
+  },
+  
+  selectDefaultImageId() {
+    // Select the current imageId by default
+    let selectedImageId = this.get("model").get("tale").get("imageId")
```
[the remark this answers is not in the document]
bodom0015: Fixed - changed pretty much all variables to `const`, since we don't need to reassign any of them

app/components/ui/tale-tab-metadata/component.js
bodom0015: Replaced double quotes with single quotes throughout

app/components/ui/tale-tab-metadata/template.hbs
bodom0015: Abstracted all inline styles to `app/styles/tale-tabs.scss` until I know of a more acceptable place to put them

app/components/ui/tale-tab-metadata/template.hbs
bodom0015: Changed readonly inputs to `<span/>`

app/components/ui/tale-tab-metadata/template.hbs
bodom0015: Changed readonly inputs to `<span/>`

app/components/ui/tale-tab-files/template.hbs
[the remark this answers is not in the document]
bodom0015: I should also mention that this tab is not currently functional, and has been hidden from view - I've simply copied this over early knowing that we will soon need such a component here to house the upcoming Tale Workspaces feature.

app/components/ui/tale-tab-metadata/component.js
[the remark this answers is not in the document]
bodom0015: Pushed an example diff of changing the `$.getJSON` to instead use the Route model here: https://github.com/whole-tale/dashboard/compare/whole-tale:refactor-view-edit-tale-metadata...refactor-tale-metadata-with-route-model

Let me know if we think this might be a workable solution, I can merge those additional changes into this PR. If this change is too big or changes too many of our assumptions, then we can discuss further before merging it.

app/components/ui/tale-tabs-selector/template.hbs
cyberdante: Also, @bodom0015, for the height I have defined Sass variables so we can keep track of the different elements that make up the app in this case:
```
        $inverted-menu-height: 48px;
        $main-menu-height: 0; // was 64px
        $wt-paddleboard-panel-height: 50px;
        $wt-main-row-height: 52px;
        $bottom-separation: 2.5rem;
        $footer-height: 94px;
        $left-panel-padding: 38px;
        $wt-paddleboard-panel-tall-height: 63px;
```
...and then I assign that panel's height by doing `height: calc(100vh - (<sum_of_other_ elements_heights_variables>))`;

cyberdante:
> Ah.. would it have worked without the `--environment=production`? I think it should have said to use `--environment=development` instead, as that seems to point to the existing API.

not specifying the environment is equivalent to `--environment=development`, and yes, it would have worked, and even compiles faster since it does not uglify the code.

app/styles/tale-tabs.scss
```diff
@@ -38,6 +38,6 @@
 }
 
 #taleMetadataTab {
-  overflow-x: auto;
-  max-height: 50vh;
+  overflow: auto;
+  height: calc(100vh - (2.5rem + 50px + 40px + 50px + 0px + 94px + 50px));
```
cyberdante: Please, use the variables instead, so we just need to change a single value and it trickles down to the whole app if a height changes for some of the other elements, e.g: `height: calc(100vh - (#{$bottom-separation} + #{$inverted-menu-height} + #{$main-menu-height} + #{$footer-height} + #{$wt-paddleboard-panel-height} + #{$wt-paddleboard-panel-tall-height}));`

cyberdante: This is the current status:
screen shot 2018-12-07 at 2 05 44 pm

app/components/ui/tale-tab-files/template.hbs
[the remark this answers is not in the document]
cyberdante: Yes @bodom0015 , no worries, I am in charge of that tab precisely. I have it assigned and in progress.

cyberdante:
@bodom0015 feel free to merge this into master.

PR #300, opened 2018-11-28
Preserve return route for non-logged-in users

bodom0015:
# Problem
When navigating to a view that requires authentication, the user is redirected to login without preserving the route they originally requested. OAuth navigates us away from the dashboard, losing all context of historically visited routes.

# Approach
I've added a bit of logic into the `AuthenticateRoute` to handle saving the originally-requested route into the `rd` query string parameter (e.g. `/login?rd=%2Fsome%2Furl%2Fencoded%2Fpath`). This way, the login route can pass this into its call to Girder's `/oauth/provider` endpoint for Girder to properly redirect us back to the requested path.

# How to Test
1. Login and navigate to one of your running instances in the dashboard - save this URL for testing (e.g. `https://dashboard.dev.wholetale.org/run/someuniquegirderid`) 
2. Open an Incognito Window and navigate to the same URL you saved above
    * You should be redirected to the `/login` route
    * The route you originally requested should be preserved in the `rd` query string parameter
3. Click the Login button to log into the WholeTale Dashboard
    * You should be logged in to the dashboard, as expected
    * You should be automatically navigated to the route specified by `rd ` previously

Caveats:
* I was unable to figure out how to strip the `?token=` query string parameter after successful authentication. This token should not be needed, as Girder's OAuth should be setting the GirderToken as a cookie, but perhaps this is not the case because `girder` and `dashboard` are served from different domains?

### DIFF
--- a/app/routes/authenticate.js
+++ b/app/routes/authenticate.js
@@ -1,8 +1,10 @@
 import Ember from 'ember';
+import { inject as service } from '@ember/service';
 
 export default Ember.Route.extend({
   tokenHandler: Ember.inject.service('token-handler'),
   userAuth: Ember.inject.service('user-auth'),
+  routerService: service('-routing'),
   currentUser: null,
   token: null,
 
@@ -28,7 +30,14 @@ export default Ember.Route.extend({
           this.get("tokenHandler").releaseWholeTaleCookie();
           this.get("userAuth").resetCurrentUser();
           router.set('currentUser', null);
-          router.transitionTo('login');
+          let returnRoute = this.routerService.router.url;
+          
+          // Prevent nexting of multiple "rd" parameters inside each other
+          if (returnRoute.indexOf("?rd=") !== -1) {
+            router.transitionTo('login');
+          } else {
+            router.transitionTo('login', { queryParams: { rd: encodeURIComponent(returnRoute) }});
+          }
           return null;
         } else {
           //   console.log("User is not null!!!");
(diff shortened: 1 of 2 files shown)


Xarthisius:
`?token=` query is indeed required due to the fact that girder and dashboard are running on different domains, and for security reasons you cannot set it for domain `b` from domain `a`.

I found an issue related to that: if you use `?rd=some_route_with%3Fquery_this%3Dfoo%26query_that%3Dbar` you end up with malformed query after login redirect: `some_route_with?query_this=foo&query_that=bar?token=...` (notice `?` is appearing twice)

## Steps to reproduce

1. `git checkout data_import` (PR #287)
1. `git merge preserve-login-request-route` (this PR)
1. build and serve on localhost:4200
1. Visit this [url](http://localhost:4200/compose?uri=https%3A%2F%2Fdev1.dataverse.org%2Fapi%2Faccess%2Fdatafile%2F34&name=My+Dataset)

## Expected behavior
Form should contain `My Dataset` in the `Title` field.

## Actual behavior
From contains `My Dataset?token`

ThomasThelen:
Nice catch Kacper I see that behavior too. It may be possible to append the old query parameters after `?token`; we would probably want to remove the first case of `?` in the original URL.

craig-willis:
@Xarthisius What would it take to set the girderToken cookie so that we don't need to pass it around on the query string?  I know we have the dashboard.wholetale.org/girder.prod.wholetale.org disconnect, but it seems like this really should just be a cookie to me.

Xarthisius:
@craig-willis @bodom0015 there's my discussion with Girder folks that my shed some light on why we went that route:
https://gitter.im/girder/girder?at=586bc175c02c1a3959e1a244

PR #328, opened 2018-12-11
Feature/tale workspace support

abrinckm:
This PR will add a new modal for selecting datasets. 

The modal will pass the selected items to the parent. The parent is responsible for what happens to the selected items. 

* You can manually test by going to Manage and clicking on the plus button next to Home.
* You can read documentation by going to app/components/ui/select-data-modal/README.md
* You can see a coding example by going to app/components/dashboard/manage/left-panel

Please note that per the conversation with Kacper and Craig the first round of implementation will only allow users to select from their data folder. There doesn't exist a context for fetching global items or workspace items.

### DIFF
--- a/app/components/dashboard/manage/left-panel/component.js
+++ b/app/components/dashboard/manage/left-panel/component.js
@@ -85,18 +85,6 @@ export default Component.extend({
     this.set("fileBreadCrumbs", state.getCurrentFileBreadcrumbs()); // new collection, reset crumbs
   },
 
-  updateSessionData(listOfDataItems) {
-    console.log(listOfDataItems);
-
-    // Add this to the dm session
-    // const store = this.get('store');
-    
-  },
-
-  cancelSelectData() {
-
-  },
-
   actions: {
     refresh() {
       let state = this.get('internalState');
--- a/app/components/dashboard/manage/left-panel/component.js
+++ b/app/components/dashboard/manage/left-panel/component.js
@@ -85,6 +85,18 @@ export default Component.extend({
     this.set("fileBreadCrumbs", state.getCurrentFileBreadcrumbs()); // new collection, reset crumbs
   },
 
+  updateSessionData(listOfDataItems) {
+    console.log(listOfDataItems);
+
+    // Add this to the dm session
+    // const store = this.get('store');
+    
+  },
+
+  cancelSelectData() {
+
+  },
+
   actions: {
     refresh() {
       let state = this.get('internalState');
@@ -350,8 +362,12 @@ export default Component.extend({
     //-----------------------------------------------------------------------------
     openRegisterModal() {
       $('.ui.modal.harvester').modal('show');
-    }
+    },
 
+    //----------------------------------------------------------------------------
+    openSelectDataModal() {
+      $('.ui.modal.selectdata').modal('show');
+    }
   }
 
 });
--- a/app/components/ui/select-data-modal/component.js
+++ b/app/components/ui/select-data-modal/component.js
@@ -10,8 +10,6 @@ export default Ember.Component.extend({
   folderNavs: service(),
   store: service(),
 
-  uuid: '33866dfb-8954-4bc9-bf61-f695b1ac2a14',
-
   selectedMenuIndex: 0,
   dataSources: A([
     O({name: 'WholeTale Catalog'}),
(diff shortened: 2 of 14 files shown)


ThomasThelen:
This is a really cool dialog; nice work!

There are a couple of behavior things that I noticed, which may be features so I'm not requesting changes on these. @craig-willis may be able to shed some light on desired behavior.

## Modal Resizes 
When navigating inside a folder the modal will re-size to fit the data which can be a little jarring
http://recordit.co/fFhWStyfOw

## Adding Duplicate Files
We're able to add duplicate files and folders
http://recordit.co/mLAsEEnplp

## Spacing
I think a small amount of padding between the areas on the left and the folders on the right would look nice (opinion).

screen shot 2018-12-11 at 10 58 40 am

abrinckm:
Thanks @ThomasThelen I'll fix those

app/components/ui/select-data-modal/component.js
```diff
@@ -0,0 +1,205 @@
...
+    return store.query('folder', {parentId, parentType, name, adapterOptions}).then(dataFolder => {
+      let dataFolderId = dataFolder.content[0].id;
+      let parentCollection = parentType;
+      self.set('rootFolderId', dataFolderId);
+      self.set('currentFolder', O({id: dataFolderId, _modelType: 'folder', parentCollection, parentId}));
+
+      return self.loadFolder.call(self, dataFolderId, 'folder');
+    }).catch(e => {
+      self.set('loading', false);
+      console.error(e);
```
ThomasThelen: If we reach this state, will any files/folders be displayed? If not we might want to show an error to the user.
Something along the lines of `Failed to load registered data. Please try again`.

abrinckm: I agree

ThomasThelen:
Awesome thanks for those changes. While testing I realized that I'm not seeing all of my registered data in Firefox (maybe it's not being able to scroll down).
screen shot 2018-12-11 at 3 11 00 pm
screen shot 2018-12-11 at 3 10 46 pm

abrinckm:
@ThomasThelen Unfortunately I'm not able to test in Firefox. The dashboard does not work in that browser. Requests made from Firefox always return `400 bad request`. See below screenshot. 

In order for me to test in Firefox, that issue will need to be fixed first.

@craig-willis This pull request cannot be resolved. Given the limited budget we have, we are close to overspending on this task. I am moving on to the other tasks.


screen shot 2018-12-11 at 8 12 35 pm

bodom0015:
@ThomasThelen @abrinckm I can confirm that Firefox does indeed run the dashboard without the 400 errors described, tested in both `61.0.2` and `63.0.3`. I suspect that your Firefox `localStorage` data may be stale, as we have seen previously with the dashboard. Would you be willing to try again in an Incognito mode?

Here is a screenshot of the later Firefox version rendering the select-data-modal:
wholetale_selectdatamodal_firefox

And here is the same view in Chrome:
wholetale_selectdatamodal_chrome

Possibly related:
Something is amiss here with the `-moz-available` style... the tag is present, but apparently Firefox does not support its use for `height` / `min-height` / `max-height`?

See the note under [`intrinsic-width`](https://caniuse.com/#feat=intrinsic-width) for Firefox, which reads: 
> Firefox does not support the height/min-height/max-height properties, only width. See [test case](https://codepen.io/shshaw/pen/Kiwaz), [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=567039).

ThomasThelen:
Thanks for checking that out @bodom0015; the new issue idea sounds good with me.

app/components/dashboard/manage/left-panel/component.js
```diff
@@ -85,6 +85,18 @@ export default Component.extend({
     this.set("fileBreadCrumbs", state.getCurrentFileBreadcrumbs()); // new collection, reset crumbs
   },
 
+  updateSessionData(listOfDataItems) {
+    console.log(listOfDataItems);
```
bodom0015: Extraneous debug statement?

abrinckm: The manage page is just a placeholder to test the modal (as per our discussion with @Xarthisius). The parent context is meant to do something with the listOfDataItems. But of course since this is just to test, manage will do nothing with the list. Just console logs for effect. Before merging this PR it would be best to remove the modal from the manage component.

abrinckm:
@bodom0015 Clearing my cache in firefox worked like a charm! There aren't anymore 400 issues.

Xarthisius:
Obosoleted by #335

PR #331, opened 2018-12-11
Handle the case when pathSuffix is falsy

### LINKED ISSUE #329: Login / logout cycle is broken with redirect handling
When user tries to login, logout then login again, they're routed to non-existing url.

Steps to reproduce:
1. Open browser in incognito mode
2. Go to https://dashboard.dev.wholetale.org and login
3. Logout
4. Login again

### Expected result

Login succeds

### Actual result

User is redirected to https://dashboard.dev.wholetale.orgundefined/?token=...

bodom0015:
# Problem
Fixes #329 

Login return route handling is flaky when logging in and out several times

# Approach
Added handling for when return route is `undefined` or falsy. Also, prevent `/login` as the return route, as this would create a cycle.

# Test Case
1. Check out and run this branch: `ember s`
2. Open an incognito tab to http://localhost:4200 and login
3. Logout
4. Login again
    * You should be logged in successfully
    * Your destination URL should **not** contain `undefined`

### DIFF
--- a/app/routes/login.js
+++ b/app/routes/login.js
@@ -15,7 +15,7 @@ export default Ember.Route.extend({
     var slashes = http.concat("//");
     var ishttp = (location.port === '') || (location.port === 80) || (location.port === 443);
     var host = slashes.concat(window.location.hostname) + (ishttp? "": ':'+location.port);
-    var pathSuffix = decodeURIComponent(params.rd) || "";
+    var pathSuffix = decodeURIComponent(params.rd || "");
     var redirectUrl = host;
     // Append return route if one exists, ignore login route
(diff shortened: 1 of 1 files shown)


Xarthisius:
@bodom0015 I tested this PR and I still see #329

craig-willis:
Walked through the test case with @bodom0015  and this was indeed happening via `ember s`. The latest commit fixes it for me.

It sounds like there is sometimes a disconnect between testing methods (`ember s` vs `ember build` versus `local.wholetale.org`).  In this case it doesn't seem to have been a factor.  My preference would be to use the `local.wholetale.org` model for testing going forward.

Xarthisius:
Tested with .local.wholetale.org, works great. Thanks @bodom0015 !

PR #335, opened 2018-12-13
Feature/tale workspace support (redo)

Xarthisius:
This PR will add a new modal for selecting datasets.

The modal will pass the selected items to the parent. The parent is responsible for what happens to the selected items.

You can manually test by going to Manage and clicking on the plus button next to Home.
You can read documentation by going to app/components/ui/select-data-modal/README.md
You can see a coding example by going to app/components/dashboard/manage/left-panel
Please note that per the conversation with Kacper and Craig the first round of implementation will only allow users to select from their data folder. There doesn't exist a context for fetching global items or workspace items.

PR #338, opened 2018-12-14
Merge test_325 into test_328 after cherry-picking from run_view_tabs

cyberdante:


Xarthisius:
Cherry-picked to #335

PR #339, opened 2018-12-18
External data tab

abrinckm:
This PR hooks up the select data modal to the external data tab and also updates the session data for the running tale when the user hits select on the modal.

This PR also adds custom rules for the /dm endpoint in the REST Adapter because the dm endpoint is not implemented in a standard RESTful way. There are 3 things to note about the `/dm` endpoint here:

1. The emberjs framework updates models using PATCH, but the dm endpoint uses PUT. So I have to override this in the adapter.
1. The model endpoints should not have nested paths. It should look like this `/dm/{id}`, where the model is `dm`. But instead the dm endpoint looks  like this `/dm/session/{id}` which forces me to have to override emberjs and insert `session` into the url.
1. Also, it is standard to send a JSON body with a request to PUT or PATCH, but the `/dm` endpoint uses query params instead. This is very non-standard. So again, I have to override the emberjs adapter to insert a query string in the url.

(*Important note*: Calling PUT `/dm/session/{id}` returns a status of 200 which is good. And GET `/dm/session/{id}` does show the files I selected to update the session. However, when I check the actual files in the running tale, I still see the original files when I first composed the tale. This is a bug on the backend? I don't know.)

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -47,8 +47,12 @@ export default DS.RESTAdapter.extend(buildQueryParamsMixin, {
     if (appendPath) {
       url += "/" + snapshot.adapterOptions.appendPath;
     }
+    let insertPath = _.get(snapshot, 'adapterOptions.insertPath');
+    if (insertPath) {
+      url = url.replace(id, `${insertPath}/${id}`);
+    }
     let queryParams = _.get(snapshot, 'adapterOptions.queryParams');
-    if (queryParams) {
+    if (queryParams) { 
       let q = this.buildQueryParams(queryParams);
       return url + "?" + q;
     }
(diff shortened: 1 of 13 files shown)


abrinckm:
To test:
1. you should be able to click on `external data` tab and see session data for the running instance. 
1. you should be able to click on the plus icon on the breadcrumb to open the select data modal.
1. in the modal, you should see on the right pane all currently mounted session data
1. in the modal, you should see on the left pane all registered datasets the user has access to
1. you should be able to add and remove items from the right pane
1. clicking select will call the backend `dm` to update the session data.

craig-willis:
@abrinckm  what is the status for the fix to the problem @bodom0015  reported?  It appears to still be a problem in this branch.

craig-willis:
Per 1/14 dev meeting, merge this and open a new issue.

PR #341, opened 2018-12-19
Tale workspace

cyberdante:
I shows some conflicts for merging but we agreed that's fine for now. Just need the work to be confirmed. Thanks!


bodom0015:
Super minor visual thing that I noticed - there is some element border that appears when you scroll down on the modal:
wholetaleworkspaceselectdataborderbug

I tried messing with the styles in Chrome's DevTools, but couldn't get it to disappear :/

craig-willis:
@cyberdante  This is another example of a PR with lots of extraneous/unrelated commits.  Talking with @Xarthisius, it may not be as straightforward to rebase as with @abrinckm's.   

Can you identify the specific commit(s) that actually belong to this PR?

A couple of other comments:
*  There is no test case.  How do we know what work was done, whether it is complete, or working as expected?
*  The PR includes a lot of whitespace changes.  What causes this?  I don't know that we see it with other developers -- maybe your editor has different defaults?

PR #343, opened 2018-12-20
Fix #342 Correct prod environment dashboard URL placeholder

### LINKED ISSUE #342: Inconsistent placeholder values for config/environments.js
See https://github.com/whole-tale/dashboard/blob/master/config/environment.js#L84-L89:
```
  if (environment === 'production') {
    ENV.apiHost = 'apiHOST';
    ENV.wholeTaleHost = 'https://dashboardHOST';
    ENV.dataOneHost = 'dataOneHOST';
    ENV.authProvider = 'authPROVIDER';
  }
```

As you can see, `dashboardHOST` is the only variable prepended with `https://`, which [deploy-dev/dashboard_local/initial_build.sh](https://github.com/whole-tale/deploy-dev/blob/dev/dashboard_local/initial_build.sh) does not handle properly

cyberdante:


Xarthisius:
Something tells me that variable is not used, since we currently have `https://https://...` on both prod and dev...

bodom0015:
Seemed to be used in precisely one place: DataONE auth in the `/run` view - see https://github.com/whole-tale/dashboard/search?q=wholeTaleHost&unscoped_q=wholeTaleHost

PR #358, opened 2019-01-18
WIP - Tale workspace operations

cyberdante:
Fixes #337. Needs rebasing before merging into master!

PR #360, opened 2019-01-22
WIP - Tale workspace operations [rebased]

### LINKED ISSUE #337: Tale workspace file/folder operations
Per discussion from https://github.com/whole-tale/dashboard/issues/304, moving workspace operations to it's own issue.

Goal:
* User can complete supported actions (listed below) in the Workspace folder for a Tale.

Supported actions:
* [x]  Upload files
* [x]  Create folders
* [x]  Rename folder/file
* [x]  Copy files from/to Home
* [x]  Move files from/to Home
* [x]  Remove folder/files
* [x] Select data from other workspaces

Completion criteria:
* [x]  User can manage files/folders using supported actions
* [ ]  Documentations updated
* [x]  Manual test case provided

Xarthisius:
Replaces #358, fixes #337.  This is a rebase of #358 that needs to be reviewed in detail to confirm that relevant commits/changes are present.

### DIFF
--- a/app/components/ui/files/destinate-folder-modal/component.js
+++ b/app/components/ui/files/destinate-folder-modal/component.js
@@ -1,9 +1,10 @@
-import Ember from 'ember';
+import Component from '@ember/component';
+import Object from '@ember/object';
+import { inject as service } from '@ember/service';
+import { A } from '@ember/array';
 import layout from './template';
 
-const service = Ember.inject.service.bind(Ember);
-
-export default Ember.Component.extend({
+export default Component.extend({
   layout,
 
   fileToMove: null,
@@ -12,19 +13,19 @@ export default Ember.Component.extend({
   userAuth: service(),
   folderNavs: service(),
 
-  folders: Ember.A(), //Array of folders in the directory
+  folders: A(), //Array of folders in the directory
 
-  destinationFolder: Ember.Object.create({}),
-  previousFolder: Ember.Object.create({}),
-  selectionTree: Ember.Object.create({}),
-  directory: Ember.Object.create({}),
+  destinationFolder: Object.create({}),
+  previousFolder: Object.create({}),
+  selectionTree: Object.create({}),
+  directory: Object.create({}),
 
   clearModal() {
-    this.set('destinationFolder', Ember.Object.create({}));
-    this.set('previousFolder', Ember.Object.create({}));
-    this.set('selectionTree', Ember.Object.create({}));
-    this.set('directory', Ember.Object.create({}));
-    this.set('folders', Ember.A());
+    this.set('destinationFolder', Object.create({}));
+    this.set('previousFolder', Object.create({}));
+    this.set('selectionTree', Object.create({}));
+    this.set('directory', Object.create({}));
+    this.set('folders', A());
     const self = this;
     const folderNavs = this.get('folderNavs');
 
@@ -114,13 +115,13 @@ export default Ember.Component.extend({
         }
         self.set('folders', folders);
       })
-      .catch(e => {
-        console.log(e);
-      })
-      .finally(() => {
-        self.set('directory', folder);
-        self.set('loading', false);
-      });
+        .catch(e => {
+          console.log(e);
+        })
+        .finally(() => {
+          self.set('directory', folder);
+          self.set('loading', false);
+        });
     },
 
     clickBack() {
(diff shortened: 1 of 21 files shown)


app/components/ui/files/destinate-folder-modal/component.js
```diff
@@ -12,19 +13,19 @@ export default Ember.Component.extend({
   userAuth: service(),
   folderNavs: service(),
 
-  folders: Ember.A(), //Array of folders in the directory
+  folders: A(), //Array of folders in the directory
```
bodom0015: Is this syntax necessary to create Arrays in Ember?

From what I understand, Ember extends raw arrays (e.g. `[]`) by default unless `EmberENV.EXTEND_PROTOTYPES` is set to false - in our case, it is not currently false. Going forward, we shouldn't need the `A([])` syntax unless that value changes.

See https://www.emberjs.com/api/ember/3.7/functions/@ember%2Farray/A

app/components/ui/files/destinate-folder-modal/component.js
```diff
@@ -1,9 +1,10 @@
-import Ember from 'ember';
+import Component from '@ember/component';
+import Object from '@ember/object';
```
bodom0015: There does not appear to be an `Object` exported from `@ember/object`, so I don't actually know which code this is importing... Is this using [`EmberObject`](https://www.emberjs.com/api/ember/3.2/classes/EmberObject) or the browser's built-in [`Object` class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object)

See https://www.emberjs.com/api/ember/3.2/modules/@ember%2Fobject

app/components/ui/files/destinate-folder-modal/component.js
```diff
@@ -12,19 +13,19 @@ export default Ember.Component.extend({
...
+  folders: A(), //Array of folders in the directory
 
-  destinationFolder: Ember.Object.create({}),
-  previousFolder: Ember.Object.create({}),
-  selectionTree: Ember.Object.create({}),
-  directory: Ember.Object.create({}),
+  destinationFolder: Object.create({}),
+  previousFolder: Object.create({}),
+  selectionTree: Object.create({}),
+  directory: Object.create({}),
```
bodom0015: This syntax seems bulky with no apparent gain in functionality. Nearly all of the methods on the classic `Object` are statically accessed, so I don't understand the value of using `Object.create({})` here when `{}` would effectively be the same and much more clear.

Am I missing something?

app/components/ui/files/destinate-folder-modal/component.js
cyberdante: Taken from [Ember documentation](https://guides.emberjs.com/v3.2.0/object-model):

> JavaScript objects don't support the observation of property value changes. Consequently, if an object is going to participate in Ember's binding system you may see an Ember.Object instead of a plain object.

> Ember.Object also provides a class system, supporting features like mixins and constructor methods. Some features in Ember's object model are not present in JavaScript classes or common patterns, but all are aligned as much as possible with the language and proposed additions.

> Ember also extends the JavaScript Array prototype with its Ember.Enumerable interface to provide change observation for arrays.

So, yes, I'd say that syntax is necessary if you want to observe the properties within the elements of any given array at some point. The price to pay for wrapping those arrays is pretty low when compared to the binding benefits that can be drawn from it.

app/components/ui/files/destinate-folder-modal/component.js
cyberdante: The answer for this would be the same as above. By wrapping objects with  `EmberObject.create({})` you gain access to the properties within these objects and can add observers to those (if needed later) instead of having to go over a refactoring at a later stage.

app/components/ui/files/destinate-folder-modal/component.js
bodom0015: I would agree if there were additional benefits over the default. I may be misinterpreting your quoted text:
> Ember also extends the JavaScript Array prototype with its Ember.Enumerable interface to provide change observation for arrays.

That last sentence indicates that we don't need the special array, as Ember will extend JavaScript's built-in `Array.prototype` to provide the same change observation. This matches what I've said above about `EXTEND_PROTOTYPES`, which we can currently leverage to simplify our syntax and not lose those benefits. I will do some more research to confirm whether this is the case and try to come up with a simple example.

It may also be worth noting that change observation and event handling can add performance overhead if it is not necessary.

app/components/ui/files/destinate-folder-modal/component.js
bodom0015: That makes sense.. While I disagree with the pattern of using them everywhere without an immediate need to observe any of the properties (e.g. `previousFolder`), the documentation on EmberObject made the benefits much more clear there.

Thank you!

cyberdante:
This contains all the necessary changes

Xarthisius:
Thanks @cyberdante !

cyberdante:
You're welcome @Xarthisius! I'll be in touch!

Issue #323, opened 2018-12-03
Run page is blank for invalid IDs


craig-willis:
When refreshing or accessing the run page by a previous url (e.g., run/instanceId) after the instance has been deleted (e.g., backend culling), the run page is blank and a 400 error appears in the console.

To repeat:
* Go to run/doesnotexist

Or
* Start an instance
* Go to the run page for the instance
* Delete the instance via API
* Refresh page 

Add error handling so that the user never sees a blank page when accessing an invalid ID.

Issue #268, opened 2018-09-14
Validate local storage state and improve error handling


craig-willis:
We've encountered yet another problem where invalid local storage state prevented users from accessing part of the system.  In this case, the "Manage" tab was failing because of a 404 on the `currentFolderId` in local storage.

This particular case can be reproduced by the following steps:
* Go to Manage > Home, create a folder called test
* Start a tale. In the tale go to home and delete the test folder
* Open Developer tools
* Go back to Manage

You'll see 400 errors for the folder, since the `currentFolderId` in local storage does not exist.

This is one instance of a problem we've seen a number of times -- the local storage state is invalid rendering parts of UI unusable with no error indication.  We need to either add validation/error handling to confirm stored IDs exist (or not rely on local storage).

cyberdante:
@craig-willis I agree that we need a better management of the `local storage`, but I was not able to reproduce this issue particularly following your steps. I even called Sebastian to look over my shoulder in case I was missing something with no results. Will try to add a mechanism to reroute from `manage.view` to `manage.index` whenever we get this type of issue. I implemented the same thing for the run page earlier this week.

craig-willis:
@cyberdante Sorry, I missed a key step in the description:

* Go to Manage > Home, create a folder called test
* **Select the test folder**
* Start a tale (e.g., Jupyter). In the tale go to `home` and delete the test folder
* Open Developer tools
* Go back to Manage

ThomasThelen:
It looks like issue #189 is caused by this, so I'm going to link it here.

craig-willis:
@bodom0015 This would be really nice to fix for the upcoming v0.5 release.  Would you be willing to look into the problem? It's pretty pervasive, but maybe we could fix some of the obvious cases quickly.

Issue #189, opened 2018-07-09
Dashboard Only Displays 'Browse`


ThomasThelen:
As mentioned on the July-9 2018 dev call this is a bug where something is getting left over in the local storage.
Workaround: Clear your cache/local storage

If you see this behavior-please post on the slack channel before clearing your cache.

<img width="1680" alt="screen shot 2018-07-09 at 12 22 29 pm" src="https://user-images.githubusercontent.com/9289652/42471854-128eeb1e-8374-11e8-8e7c-9f4a7e338bfd.png">

abrinckm:
If using Chrome please open dev tools -> Application -> Storage -> Local Storage -> localhost and copy paste all the information you see there. Thank you!

ThomasThelen:
Closing this since we're already tracking the app storage issues.

Issue #185, opened 2018-07-03
'Run' tab becomes unresponsive after shutting down an instance


Xarthisius:
Step to reproduce:
1. Login and launch a Tale (from 'browse' view)
2. Go to 'Run', select an Instance from right panel, delete instance
3. Go to different tab, then try to go back to 'Run'

This traceback might be relevant:

```
vendor-4c400ad00abf6b6ac14145a470b715f2.js:1859 Error while processing route: run.view Attempted to handle event `loadingData` on <instance:5b3b7bb08b0b4b0001c0189f> while in state root.deleted.saved. Called with {_id: 2957, _label: Fetching instance' with id: 5b3b7bb08b0b4b0001c0189f, _state: undefined, _result: undefined, _subscribers: }. Error: Attempted to handle event `loadingData` on <instance:5b3b7bb08b0b4b0001c0189f> while in state root.deleted.saved. Called with {_id: 2957, _label: Fetching instance' with id: 5b3b7bb08b0b4b0001c0189f, _state: undefined, _result: undefined, _subscribers: }.
    at new n (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:1866:13)
    at e._unhandledEvent (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10015:121)
    at e.send (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10004:19)
    at e.loadingData (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10001:253)
    at o._scheduleFetch (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10165:10)
    at o._findRecord (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10155:190)
    at o.findRecord (https://newdashboard.dev.wholetale.org/assets/vendor-4c400ad00abf6b6ac14145a470b715f2.js:10155:49)
    at o.model (https://newdashboard.dev.wholetale.org/assets/wholetale-264c0f35f30555a73a1c04a0621bc4be.js:568:1255)
[...]

PR #376, opened 2019-02-04
[WIP] Fixing local storage issues.

Xarthisius:
Multiple times over the lifetime of the project, we've encountered issues with `localStorage` being out of date. It caused a lot of issues: (#323, #268, #189, #185...)

### Approach
Before returning any object ID from `localStorage` verify that it exists on a backend. Return `null` otherwise.


Xarthisius:
See #326 and #378

Issue #168, opened 2018-06-27
Tale Browser Has Incorrect State after Tale Deletion in Search


ThomasThelen:
I was deleting a set of tales whose name shared a common word. To do this I went to the `Browse` tab and searched the tales for the common word. After deleting one, the widget reset the search, but left the search term in the search field (see image).

Build: `56d4af1`

To Reproduce:
   1. Create two tales, name them 'mytale1' and 'mytale2`
   2. Go to `Browse`
   3. Search for `myTale`
   4. Delete one of them
   5. Note the behavior described above.

<img width="1680" alt="screen shot 2018-06-27 at 2 40 00 pm" src="https://user-images.githubusercontent.com/9289652/42001262-558aecac-7a18-11e8-8bbd-ee657c85acb4.png">

ThomasThelen:
Update: Still present in build `5eae564` (Nov-15-2018)

Xarthisius:
Fixed by https://github.com/whole-tale/dashboard/pull/393

PR #393, opened 2019-02-15
Refilter Tale list after deleting a Tale

### LINKED ISSUE #390: Browse filter state mismatch when deleting a Tale
### Steps to Reproduce
1. Compose a Tale
2. Navigate to "Browse" and filter to "Mine"
3. Delete a Tale

### Expected Results
Tale should be deleted, and one of two things should happen:
* Either the filter text should be reset to "All"
* Or the list should actually stay filtered to "Mine"

### Actual Results
* Tale is deleted
* List is refiltered to show all Tale
* Filter at the top (incorrectly) shows "Mine"

bodom0015:
### Problem
Fixes #390 

### Approach
Just needed to call `setFilter()` after resyncing Tale data

### How to Test
Prerequisite: at least one Tale has been created in the system that does not belong to you

1. Compose a Tale
2. Navigate to "Browse" and filter to "Mine"
    * You should see that Tales you do not own are hidden
3. Delete the Tale that you own
    * You should see that Tales you do not own are still hidden

### DIFF
--- a/app/components/ui/tale-browser/component.js
+++ b/app/components/ui/tale-browser/component.js
@@ -233,6 +233,7 @@ export default Ember.Component.extend({
         // refresh
         component.updateModels(component, component.get('models'));
         component.set('selectedTale', undefined);
+        component.setFilter();
       });
 
       return false;


ThomasThelen:
Just tested this and it looks good. This also fixes #168

Xarthisius:
Works as advertised (I tested #168 too). Thanks!

PR #396, opened 2019-02-15
Refilter with myData when calling goBack in select-data-modal

### LINKED ISSUE #389: My Data displays all catalog data after navigating up
A bug report and a change request.

To repeat:
* Start a tale instance
* Select Files > External Data > +  to open the Select Data modal
* Select My Data and browse into  a dataset
* Select "<-" to return to the My Data
* Note that the full catalog is displayed.

Also, I think we should reverse the order of WT Catalog/My Data -- MyData should be at the top and selected by default.

bodom0015:
### Problem
Fixes #389 

### Approach
We were not passing the `myData` parameter to `GET /dataset` in this case - pass this parameter when calling `goBack()`

### How to Test
2. Launch a Tale
3. Navigate to Run > Files > External Data > +  to open the Select Data modal
4. Select My Data
    * You should see the list is filtered to show only datasets that you (the user) have registered
5. Browse into a dataset
6. Select "<-" to return to the My Data
    * You should see the list still shows only datasets that you (the user) have registered

### DIFF
--- a/app/components/ui/select-data-modal/component.js
+++ b/app/components/ui/select-data-modal/component.js
@@ -186,14 +186,7 @@ export default Component.extend({
 
         const self = this;
         if (parentId == null || parentId == this.get('rootFolderId')) {
-            return store.query('dataset', adapterOptions).then(datasets => {
-                let catalogId = this.get('rootFolderId');
-                self.set('loading', false);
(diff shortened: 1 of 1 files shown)


Xarthisius:
LGTM!

PR #406, opened 2019-02-15
Disable Run > Files > Tale Workspaces UI if insufficient Tale access

### LINKED ISSUE #391: UI should disable Tale Workspaces features for users with insufficient Tale access
There are several places that a user can see Tales listed to which they don't have access.
* Browse: user can currently "Launch" an Instance of another user's Tale
* Run: user can then view a "read-only" Instance from the Tale

This presents a problem when the user attempts to view or add items to the Workspace for the Tale:
* Run > Files > Tale Workspaces: the user expects to be able to view the contents of the workspace
* Run > Files > Tale Workspaces > (+) button: the user expects to be able to import data from other workspaces, or to create/upload new folders or items

Since the user can't read from or write to this Tale's Workspace, the UI should prevent them from attempting to do so. This could be done by:
* Somehow translate/detect access level of Tale in the UI (is this exposed? We will need it for copy on launch anyways)
* Disabling the Run > Files > Tale Workspaces nav when user doesn't have read access
* Disabling the Run > Files > Tale Workspaces > (+) button when the user does not have write access

NOTE: Copy on Launch may remove the need for such a fix

bodom0015:
### Problem
User can launch a Tale to which they don't have sufficient read or write access. Tale Workspaces UI becomes non-functional in this case, since we can't read the Workspace or upload files/folders to it.

Fixes #391

### Approach
Disallow user from selecting Tale Workspaces without read access, disallow Add button without write access

### How to Test
3. Launch a Tale that does not allow you read access
4. Navigate to Run > Files
    * You should see that Tale Workspaces has been greyed out, disallowing reads
5. Launch a Tale that allows you to read, but not write
6. Navigate to Run > Files > Tale Workspaces
    * You should see the Tale's Workspace folders/files are listed here
    * You should see that (+) has been greyed out, disallowing writes

### DIFF
--- a/app/components/ui/files/bread-crumbs/template.hbs
+++ b/app/components/ui/files/bread-crumbs/template.hbs
@@ -2,7 +2,12 @@
     <div class="ui large breadcrumb">
         <a class="section {{if justDescription 'bold'}}" {{action "navClicked" currentNav}}> {{#if justDescription}} {{currentNav.displayName}} {{else}} {{currentNav.name}} {{/if}}</a>
         {{#if justDescription}}
-            <i class="fas fa-plus-circle icon blue large" onclick={{action 'triggerBreadcrumbAction' currentNav.name}}></i>
+            {{#if (and (lt model.tale._accessLevel 1) (eq currentNav.name 'workspace'))}}
+                <i class="fas fa-plus-circle icon blue large disabled" disabled style="cursor:not-allowed !important;" 
+                    data-tooltip="You do not have write access to this Tale." data-inverted=""></i>
+            {{else}}
+                <i class="fas fa-plus-circle icon blue large" onclick={{action 'triggerBreadcrumbAction' currentNav.name}}></i>
+            {{/if}}
             {{#if displayFoldersMenu}}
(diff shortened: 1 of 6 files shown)


craig-willis:
Per 2/18 discussion, when testing with https://github.com/whole-tale/girder_wholetale/pull/243, the underlying problem fixed here is unlikely, but the PR is still valid.  More important is correct handling of the read-only case (including public=true).  If the Tale/Workspace is read-only, the user should be able to view the contents of the workspace, download and copy items, but not write, upload, rename, create folder, remove, or move, etc.

bodom0015:
I've updated the PR to handle the "Read-only Tale" case a bit better:
* Ensured that the Tale Workspaces nav is enabled based on both `accessLevel` and `public`
* Ensured that the (+) button is disabled appropriately
* Removed all folder/file dropdown options except for "Copy to Home" and "Download"

Xarthisius:
LGTM! I found one issue, but I don't know if it should be addressed in this PR (or at all):
It's possible to try to move files from ro workspaces which results in error shown below:
workspace

Do we want to add logic checking if source workspace is ro and gray out "Move to Workspace" button?

PR #408, opened 2019-02-18
Hide 'Input Data' section on Compose view. Fixes #407

### LINKED ISSUE #407: Remove/change Compose > Input Data placeholder
Per https://github.com/whole-tale/dashboard/issues/362#issuecomment-458606206 until #366 is fixed, on Compose page we should change or remove: "Select from Data Panel" placeholder in "Input Data:" field.

bodom0015:
### Problem
'Input Data' portion of Compose view is no longer functional

### Approach
Comment it out for now, until we can add back this functionality

### How to Test
2. Login and navigate to Compose
    * You should no longer see "Input Data" or "Select from Data Panel" on the form

### DIFF
--- a/app/templates/forms/new-tale-input.hbs
+++ b/app/templates/forms/new-tale-input.hbs
@@ -24,7 +24,6 @@
             </div>
         {{/if}}
     </div>
-    {{!--
     <div class="row panel container">
         <div class="five wide column field">
             <label>Input data:</label>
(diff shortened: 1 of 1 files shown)


Xarthisius:
Per 2/18 discussion, in case of "Analyze in WT" we want the `Input field` to be present. Only text/placeholder should be removed in this PR.

PR #412, opened 2019-02-19
Turn off backgroundReload when deleting a Tale in the legacy Tale View page

### LINKED ISSUE #395: Deleting Tale from View Page Doesn't Redirect
Tested in Firefox and Chrome.

When deleting a tale from the view page, the Tale will delete, but the user isn't re-directed to the browse page.

I saw the following error in the console
```
vendor-3f5071e34f3d87102452624f03ecb03b.js:1901 Error while processing route: browse Attempted to handle event `pushedData` on <tale:5c66055b57d5420001478423> while in state root.deleted.saved.  
Error: Attempted to handle event `pushedData` on <tale:5c66055b57d5420001478423> while in state root.deleted.saved. 
```

bodom0015:
### Problem
When deleting a Tale from the legacy "Tale View" page, a stack trace occurs preventing the user from being redirected properly to the "Browse" page. This occurs because EmberJS is attempting to automatically fetch a record while it is being deleted or shortly thereafter.

Fixes #395

### Approach
When fetching records, EmberJS has a behavior that will automatically cache them for later updates. You can pass an optional parameter to disable this behavior, which is encouraged when deleting a model.

> The backgroundReload option is used to prevent the fetching of the destroyed record, since findRecord() automatically schedules a fetch of the record from the adapter.

Since I was unsure where the current model being passed in was actually being fetched, or what other views might be reusing the same model, the least intrusive fix seemed to be to add an additional fetch here that explicitly specifies `backgroundReload: false`.

See https://guides.emberjs.com/release/models/creating-updating-and-deleting-records/#toc_deleting-records

### How to Test
2. Login to the Dashboard
3. Compose a Tale, if you haven't already
4. Stop/delete the Tale's Instance, if you haven't already
5. Visit the "Tale View" page for this Tale
6. At the top-right, click "Delete Tale"
    * You should be prompted for confirmation
7. Confirm the deletion
    * You should be redirected to the "Browse" page
    * Your deleted Tale should no longer appear in the list here
    * You shouldn't see a big angry stack trace in the logs

### DIFF
--- a/app/controllers/tale/view.js
+++ b/app/controllers/tale/view.js
@@ -260,11 +260,15 @@ export default Controller.extend({
     approveDelete(model) {
       let component = this;
 
-      model.destroyRecord({
-        reload: true
-      }).then(function () {
-        component.transitionToRoute('browse');
+      let taleId = model.get('_id');
+      this.get('store').findRecord('tale', taleId, { backgroundReload: false }).then(function(tale) {
+        tale.destroyRecord({
+          reload: false
+        }).then(function () {
+          component.transitionToRoute('browse');
+        });
       });
+      
 
       return false;
     },
--- a/app/routes/browse.js
+++ b/app/routes/browse.js
@@ -14,7 +14,8 @@ export default AuthenticateRoute.extend({
         }
       }),
       tales: this.get('store').findAll('tale', {
-        reload: true,
+        reload: false,
+        backgroundReload: false,
         adapterOptions: {
           queryParams: {
             sort: "created",
--- a/app/routes/tale/index.js
+++ b/app/routes/tale/index.js
@@ -4,7 +4,8 @@ export default AuthenticateRoute.extend({
   model() {
     return {
       tales: this.get('store').findAll('tale', {
-        reload: true,
+        backgroundReload: false,
+        reload: false,
         adapterOptions: {
           queryParams: {
             sort: "created",


Xarthisius:
I get an angry stack when I follow the test case:

```
ember-console.js:29 Error while processing route: browse Attempted to handle event `pushedData` on <tale:5c6d89eb4c68440001837909> while in state root.deleted.saved.  Error: Attempted to handle event `pushedData` on <tale:5c6d89eb4c68440001837909> while in state root.deleted.saved. 
    at new EmberError (https://dashboard.local.wholetale.org/assets/vendor.js:29038:25)
    at InternalModel._unhandledEvent (https://dashboard.local.wholetale.org/assets/vendor.js:232560:11)
    at InternalModel.send (https://dashboard.local.wholetale.org/assets/vendor.js:232434:12)
    at InternalModel.pushedData (https://dashboard.local.wholetale.org/assets/vendor.js:232348:10)
    at InternalModel.setupData (https://dashboard.local.wholetale.org/assets/vendor.js:232295:10)
    at Class._load (https://dashboard.local.wholetale.org/assets/vendor.js:236854:19)
    at Class._pushInternalModel (https://dashboard.local.wholetale.org/assets/vendor.js:237217:30)
    at internalModelOrModels._backburner.join (https://dashboard.local.wholetale.org/assets/vendor.js:237157:36)
    at Backburner._run (https://dashboard.local.wholetale.org/assets/vendor.js:25440:35)
    at Backburner._join (https://dashboard.local.wholetale.org/assets/vendor.js:25418:29)
error @ ember-console.js:29
logError @ router.js:925
error @ router.js:867
triggerEvent @ router.js:1024
_trigger @ router.js:89
trigger @ router.js:507
(anonymous) @ router.js:429
tryCatcher @ rsvp.js:200
invokeCallback @ rsvp.js:372
publish @ rsvp.js:358
publishRejection @ rsvp.js:293
(anonymous) @ rsvp.js:14
invoke @ backburner.js:205
flush @ backburner.js:125
flush @ backburner.js:278
end @ backburner.js:410
_run @ backburner.js:760
_join @ backburner.js:736
join @ backburner.js:477
run.join @ ember-metal.js:456
hash.success @ rest.js:960
fire @ jquery.js:3268
fireWith @ jquery.js:3398
done @ jquery.js:9305
(anonymous) @ jquery.js:9548
load (async)
send @ jquery.js:9567
ajax @ jquery.js:9206
_ajaxRequest @ rest.js:982
(anonymous) @ rest.js:972
initializePromise @ rsvp.js:397
Promise @ rsvp.js:877
ajax @ rest.js:955
findAll @ rest.js:471
_findAll @ -private.js:8786
_fetchAll @ -private.js:11251
findAll @ -private.js:11229
model @ browse.js:19
superWrapper @ ember-utils.js:428
deserialize @ route.js:951
applyHook @ router.js:211
runSharedModelHook @ router.js:662
getModel @ router.js:896
tryCatcher @ rsvp.js:200
invokeCallback @ rsvp.js:372
publish @ rsvp.js:358
(anonymous) @ rsvp.js:14
invoke @ backburner.js:205
flush @ backburner.js:125
flush @ backburner.js:278
end @ backburner.js:410
_run @ backburner.js:760
_join @ backburner.js:736
join @ backburner.js:477
run.join @ ember-metal.js:456
(anonymous) @ ember-glimmer.js:974
exports.flaggedInstrument @ ember-metal.js:4582
(anonymous) @ ember-glimmer.js:973
XHR finished loading: GET "https://girder.local.wholetale.org/api/v1/tale?sort=created&sortdir=1&limit=0".
rsvp.js:27 Uncaught EmberError {stack: "Error: Attempted to handle event `pushedData` on <…rd.local.wholetale.org/assets/vendor.js:25418:29)", description: undefined, fileName: undefined, lineNumber: undefined, message: "Attempted to handle event `pushedData` on <tale:5c…440001837909> while in state root.deleted.saved. ", …}
```

which I  think is the same traceback I get on current master.

bodom0015:
In my initial testing, it seemed like this had been fixed... but as @Xarthisius has said, we're still seeing this pop up intermittently.

More investigation is needed into what's happening here, but I suspect there's some race condition or an underlying Ember resource is somwhoe still attempting to auto-reload itself behind-the-scenes.

bodom0015:
Pushed another attempted fix... We seem to be pulling from the `/tale` endpoint in several different spots. This should be the one that feeds them to the legacy `tale.view` page.

With other frameworks, it is typical to have a "service" for each model type that handles the lifecycles and synchronization of those objects across different views. I think that some of this is handled by `ember-data` / `DS.store`, but there are certainly a few duplicate queries still around.

craig-willis:
I seem to be getting the same error as @xarthisius with the latest changes:
```
vendor-3fec92aafa2a46ab1024fb73ecea7c3c.js:1860 Error while processing route: browse Attempted to handle event `pushedData` on <tale:5c7e90bc27184e0001890cce> while in state root.deleted.saved.  Error: Attempted to handle event `pushedData` on <tale:5c7e90bc27184e0001890cce> while in state root.deleted.saved.
```

And it doesn't redirect to the browse page.

@ThomasThelen -- could you re-test this and report if you see the same?

ThomasThelen:
Can confirm that on Firefox and Chrome I'm not getting a redirect and that I'm receiving the same errors. The Tale _does_ get deleted.

bodom0015:
I've pushed one more fix for this.... I have maybe an idea of why we keep seeing this and then not under different circumstances.

My only possible explanation: the models that are cached and how many copies there are depends on which views you have already visited during your session. That is, if you (like myself) go directly to `/run` when testing then you may not hit the error, but if you navigate from `/browse` (which caches a copy of the models from `/tale`, and *then* navigate to run then you will see the bug. This is because the Browse view/route actually pulls a reference to all of the Tales, while the Run view does not.

It is still unclear if this invalid data in the cache survives a reload of the tab, but I suspect that it does not.

craig-willis:
We're not going to do this since the old view is going away.

PR #413, opened 2019-02-19
Remove leading slash from external data

### LINKED ISSUE #411: Run > Files > External Data: items listed with a leading slash
Files in the external data view have leading slashes:

data

### Expected result
No leading slashes

### How to reproduce
1. Deploy locally and register Ligo Tale
1. Run Ligo Tale
1. Navigate to Run > Files > External Data

bodom0015:
### Problem
Sometimes External Data displays datasets prepended with a leading slash.

Fixes #411 

### Approach
Replace `/` with an empty string when building up the `mountPath`, as we do with other places in that view (such as `allSelectedItems` in the `select-data-modal`).

NOTE: I am just following the established pattern here, but I wonder if the `g` at the end of the replace regex could cause problems with dataSets that should actually contain such slashes? For example: https://search.dataone.org/view/doi:10.5063/F1C827JF, whose title contains a URL/DOI, display as `Data supporting meta‐analysis of multiple ecosystem outcomes from organic amendment additions to rangelands (https:doi.org10.1111gcb.14535)` - this is obviously not ideal

### How to Test
1. Checkout and run this branch locally, rebuild dashboard
2. Register a dataset known to display this problematic behavior (e.g. LIGO Tale)
3. Login and launch a Tale, if you haven't already
4. Navigate to Run > Files > External Data
5. Open the "Select Data", add the dataset to the Tale, and click "Select"
    * You should see the modal dismiss itself
    * You should see the dataset appear in the External Data list
    * You should see that the dataset name is no longer lead by slashes

### DIFF
--- a/app/components/ui/tale-tab-files/component.js
+++ b/app/components/ui/tale-tab-files/component.js
@@ -271,15 +271,14 @@ export default Component.extend({
                     'resources': payload
                 });
             } else if (nav.command === "user_data") {
-
               let taleId = this.get('model.taleId'); // state.currentInstanceId;
-                let taleDatasetContents = controller.get('store').findRecord('tale', taleId)
-                    .then(tale => {
-                        return tale.get('dataSet').map(dataset => {
-                            let { itemId, mountPath, _modelType } = dataset;
-                            return { id: itemId, name: mountPath, _modelType };
-                        })
-                    });
+              let taleDatasetContents = controller.get('store').findRecord('tale', taleId, { reload: false })
+                .then(tale => {
+                    return tale.get('dataSet').map(dataset => {
+                        let { itemId, mountPath, _modelType } = dataset;
+                        return { id: itemId, name: mountPath, _modelType };
+                    })
+                });
               itemContents = Promise.resolve(A([]));
               folderContents = taleDatasetContents.then(_taleDatasetContents => {
                 newModel.sessionContents = _taleDatasetContents;
@@ -471,7 +470,7 @@ export default Component.extend({
             // Build up our dataSet list
             let dataSet = listOfSelectedItems.map(item => {
                 let {id, name, _modelType} = item;
-                return {itemId: id, mountPath: name, _modelType};
+                return {itemId: id, mountPath: name.replace(/\//g, ''), _modelType};
             });
             this.session.set('dataSet', dataSet);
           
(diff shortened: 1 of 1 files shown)


Xarthisius:
> NOTE: I am just following the established pattern here, but I wonder if the `g` at the end of the replace regex could cause problems with dataSets that should actually contain such slashes? For example: https://search.dataone.org/view/doi:10.5063/F1C827JF, whose title contains a URL/DOI, display as `Data supporting meta‐analysis of multiple ecosystem outcomes from organic amendment additions to rangelands (https:doi.org10.1111gcb.14535)` - this is obviously not ideal

Yeah, I'd definitely remove only leading (possibly trailing, but that should happen on the backend already) slash. Something like `replace(/^\/|\/$/g, '')`?

bodom0015:
I think your proposed regex should work for leading/trailing slashes. I've added a small bit of handling to the UI, since we are already looping through the dataSets anyway to map them

Xarthisius:
> I think your proposed regex should work for leading/trailing slashes, but if the fix is on the API side, then shouldn't this issue (#411) have been filed under `girder_wholetale`?

I only meant that for trailing slash. Leading slash is expected and required on the backend.

bodom0015:
I'm ignoring trailing slashes atm, but could strip those off as well if we think it might be worth it.

In the meantime, I've gone with the following:
```javascript
let name = mountPath[0] === '/' ? mountPath.substring(1) : mountPath;
```

If `name` starts with a slash, we grab everything but the leading slash. Otherwise we grab the whole thing.
______________

#### Aside: Am I doing it right?

When I send the payload for updating the "Selected Datasets" for the Tale, I am sending the following: https://github.com/whole-tale/dashboard/pull/413/files#diff-b22ebb3f6d2f7b177bc2c4fbb882b304R479:
```javascript
let mountPath = name[0] === '/' ? name.substring(1) : name;
```

Does the `mountPath` need to preserve the leading slash in this case? Or is it safe to strip it here as well?

The more questions that I ask, the larger my sense that the backend should really be handling the normalization from the `mountPath` (possible leading slash) -> `name` (no leading slash) when returning from the `GET /dataset` endpoint.

What do you think, @Xarthisius?

Xarthisius:
@bodom0015 `mountPath` requires a leading slash. It's a path relative to the root of the mount point.

I'm not sure I understand what you mean by handling normalization on the backend. One thing I'd like to stress is that resource `name` and `mountPath` are completely separate concepts. We currently default to `mountPath` == `'/' + name`, but that's just a choice. Explicit `mountPath` was implemented to allow users to override not human friendly dataset names, such as:

```
data/Replication Data for: The Economic Consequences of Partisanship in a Polarized Era/
```
into something more keyboard friendly e.g. `data/dataset01/`. As a side effect we need to implement that '/' stripping logic on the UI side unfortunately.

Xarthisius:
After discussion on 02/20, I'm closing this PR cause the test case represented something that cannot be done on the UI side.

Issue #381, opened 2019-02-08
Strange reload behavior


craig-willis:
I'm seeing  cases where moving between tabs in the UI causes the page to fully reload and take me back to the original page I was on.  This seems to happen when I transition to a new tab before the first has completely loaded. 

For example:
* Launch a Tale. Select Run tab. While interact is loading select Files.  Page reloads and taken back to interact Tab.
* Select Run then Manage before page loads. Taken back to Run with reload.

I can't repeat with either the browse or compose pages.

bodom0015:
Another repeatable cases (may be the same case as your second point above):
1. Navigate directly to Run
2. Click Files, then choose External Data (if it is not already chosen)
    * If External Data was not already chosen and is clicked, page reloads here
3. Switch to Manage
4. Switch back to Run and click Files again
    * Page reloads here, since External Data is selected on init
    * NOTE: This does not seem to happen when "Home" or "Tale Workspaces" are loaded on init

This leads me to believe that this bug has something to do with the init logic on the Run > Files > External Data tab.

bodom0015:
This is caused by my hacky fix for the `internal-state` stuff.

What happens:
1. `internal-state` attempts to check currentFolderID, which for some reason is `undefined`
2. My explicit short-circuiting checks for `undefined` are somehow failing
3. getCurrentFolderID incorrectly receives an error when looking up `currentFolderID=undefined`, and refreshes the page (correctly) to fix the bad state

There appears to be some sort of race condition or initialization error here that prevents the first value of currentFolderID to be set or detected properly.

More investigation is needed.

bodom0015:
So here's the trick: Manage and Run > Files > External Data no longer function as they used to. All of the "Files / Folders" views assumed that there would always be a currentFolderID. Now with "External Data" reading from the Tale and "Manage > Data" reading from `/dataset`, there is a disconnect.

Some of our views are still assuming that without a `currentFolderID`, we are in an error state and need to reload.

ThomasThelen:
I tested a change where the call to retrieve the currentFolderID returns the user's ID on error and didn't run across any issues-but I'm not sure how it didn't break anything so I'm going to test a little more before PRing it.

bodom0015:
Test case is no longer valid, since there is no longer a Run tab.

Is this issue still relevant?

Xarthisius:
> Test case is no longer valid, since there is no longer a Run tab.
> 
> Is this issue still relevant?

Nope.

PR #414, opened 2019-02-19
Return the home folder if the currentFolderID is empty/undefined

ThomasThelen:
I added a method to the folder-navs to return the ID of the home folder which is used in the internal-state service. 

If the `currentFolderID` is undefined, then return the ID of the home folder. This should fix issue #381.

To Test:
1. Clone & Deploy locally
2. Delete the `currentFolderID` local storage object
3. Browse WT
4. Repeat 2-3

The important pages to view are the compose and run pages.

### DIFF
--- a/app/services/folder-navs.js
+++ b/app/services/folder-navs.js
@@ -6,7 +6,8 @@ import {
 export default Service.extend({
     userAuth: service(),
     internalState: service(),
-
+    store: service(),
+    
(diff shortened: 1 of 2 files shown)


ThomasThelen:
The test case I gave was to make sure I didn't regress the original fix for the currentFolderID where some pages would show up as blue.

PR #424, opened 2019-02-21
Handle 'undefined' string literal in internal-state null-checking

### LINKED ISSUE #422: Manage > Data local storage problem in v0.6.rc2
I'm seeing two issues on the v0.6rc2 manage data page:
1. Registered dataset doesn't appear without refresh
2. Local storage currentFolderId is undefined, resulting in 400 errors

Test case:
1. https://dashboard.stage.wholetale.org
2. Manage > Data remove any existing registered datasets
3. Register dataset doi:10.5065/D6TB1525
4. Note: Dataset doesn't appear in folder after registration
5. Select Home then Data. Dataset should appear
6. Select Run then Select Manage

```
GET https://girder.stage.wholetale.org/api/v1/folder/undefined 400 (Bad Request)
girder.stage.wholetale.org/api/v1/item?limit=0&folderId=undefined&reload=true:1 GET https://girder.stage.wholetale.org/api/v1/item?limit=0&folderId=undefined&reload=true 400 (Bad Request)
girder.stage.wholetale.org/api/v1/folder?limit=0&parentId=undefined&parentType=folder&reload=true:1 GET https://girder.stage.wholetale.org/api/v1/folder?limit=0&parentId=undefined&parentType=folder&reload=true 400 (Bad Request)
```

Under local storage, there is no `currentFolderId` key.

bodom0015:
### Problem
While our `internal-state` service does handle `undefined` values properly, it does not handle the `"undefined"` string literal.

Fixes #422

### Approach
While the thorough/correct fix would likely be to make sure we aren't assigning the `"undefined"` string literal anywhere (as this is a meaningless arbitrary string), to get things moving in the short-term we can simply handle this as a special value.

Note that this anti-pattern only currently doesn't present many problems because the field represents a specially-formatted ID - the same pattern for null-checking could not be applied to other fields, such as `currentFolderName`, since `"undefined"` is actually a valid name for a user to give to a folder

### How to Test
3. Navigate to Manage > Data
4. Register a dataset, if you haven't already
    * Your view should automatically update or refresh to display the new dataset
5. Delete the dataset you just registered
    * Your view should automatically update to show that the dataset has been disassociated from your user
6. Navigate to Run
7. Navigate back to Manage
    * You should be able to navigate successfully

Compare with https://dashboard.stage.wholetale.org/manage for reference

### DIFF
--- a/app/components/dashboard/manage/left-panel/component.js
+++ b/app/components/dashboard/manage/left-panel/component.js
@@ -92,12 +92,23 @@ export default Component.extend({
 
     this.set('fileBreadCrumbs', state.getCurrentFileBreadcrumbs()); // new collection, reset crumbs
   },
+  resync() {
+      let nav = this.get('currentNav');
+      this.send('navClicked', nav);
+  },
 
   actions: {
     refresh(adapterOptions = { queryParams: { limit: '0' } }) {
       let state = this.get('internalState');
       let myController = this;
       let itemID = state.getCurrentFolderID();
+      
+      // Short-circuit: Datasets are no longer read from these endpoints
+      let nav = this.get('currentNav');
+      if (nav.command == 'user_data') {
+        myController.resync();
+        return;
+      }
 
       let folderContents = myController.store.query('folder', {
         parentId: itemID,
--- a/app/components/ui/tale-tab-files/component.js
+++ b/app/components/ui/tale-tab-files/component.js
@@ -126,6 +126,13 @@ export default Component.extend({
             let state = this.get('internalState');
             let myController = this;
             let itemID = state.getCurrentFolderID();
+      
+            // Short-circuit: Datasets are no longer read from these endpoints
+            let nav = this.get('currentNav');
+            if (nav.command == 'user_data') {
+                myController.resync();
+                return;
+            }
 
             let folderContents = myController.store.query('folder', {
                 parentId: itemID,
--- a/app/services/internal-state.js
+++ b/app/services/internal-state.js
@@ -39,7 +39,7 @@ export default Service.extend({
       let self=this;
 
       // Skip checking falsey ID strings
-      if (typeof(currentFolderID) !== "undefined" && currentFolderID) {
+      if (typeof(currentFolderID) !== "undefined" && currentFolderID !== "undefined" && currentFolderID) {
         self.store.findRecord('folder', currentFolderID).then(function(result) {
           // folder exists - effectively a noop
           // NOTE: after first lookup, this result is cached in the store


craig-willis:
Testing this on my local deploy, I get two 400s at step 4:

```
GET https://girder.local.wholetale.org/api/v1/item?limit=0&reload=true (400 bad request)
GET https://girder.local.wholetale.org/api/v1/folder?limit=0&parentType=folder&reload=true (400 bad request)
```

bodom0015:
Good catch, guys! This has certainly been a hard one to track down...

For me it does not seem to be a symptom of actually registering a dataset, as the following works for me every time
1. Refresh the page
2. Register a new dataset (e.g. one that does not already appear in the view)
3. Without refreshing, register another new dataset
4. Still without refreshing, register a third new dataset

What does **not** work properly is the following:
1. Refresh the page
2. Delete a dataset (before registering a new one)
3. Register a new dataset (e.g. one that does not already appear in the view)
    * You'll see the 400 errors in the console
    * The view won't update properly, and newly registered data is never shown
4. Click the Home nav, then click back to Data
    * Dataset list clears, but never displays the list
    * No error messages appear in the console
    * Only a hard refresh can correct this state - navigating to Run and back or changing navs does not correct the issue

Closest I can tell so far, something outside of the folder nav service is still not handling undefined properly. Those 400s are coming back from `GET /api/folder`, which appears to be coming from the `refresh()` function inside the `manage/left-panel` component.

More investigation is needed to figure out exactly what is happening there

Xarthisius:
@bodom0015 do you want to handle the latter case as a part of this PR, or is it good to merge?

bodom0015:
@Xarthisius here is the status of the issues present here:
1. As far as I know the original issue (#422) is a huge problem that is addressed by this PR
2. I am now seeing and can address the issue reported by @craig-willis (about getting 400s when registering a dataset)
3. I am unable to replicate the issue reported by @ThomasThelen (about getting 400s when deleting a dataset) 

@craig-willis the view seemed completely unusable before due to the instability, right? If this PR has been tested and is considered an improvement over the current behavior, then we should merge it up and address other issues as they come up

bodom0015:
I went ahead and pushed a small fix that should get rid of the 400s when registering data - there is logic there that is still trying to call the `/folder` and `/item` endpoints. I believe this error was innocuous, but was preventing the notification popups from appearing in the success case. I can file a new issue for the strange edge case I've reported above. I am hoping that it's more of an issue for testers ,since users probably wouldn't just be sitting there registering and deleting over and over again.

@ThomasThelen if you are able to reproduce the behavior you mentioned above, would you be willing to file it as a new issue? I have not been able to reproduce an error on delete

Issue #305, opened 2018-11-29
Update publish to use workspaces


craig-willis:
Per the [design discussions](https://docs.google.com/document/d/1pl9FwVNeXyVVTccXLoTIeXsViWflbPpoQ5DLY44g5cw/edit#heading=h.y3lmftolmepl), publishing will operate on the Tale workspace and no longer allow the user to select data from Home. Note: there is still some discussion about whether to allow the user to select items to be published, so for now we will not implement this.

Goal: 
* During publish, the contents of the current Tale workspace are published. 

Completion criteria:
* [x] User can publish tale based on contents of current tale workspace
* [x] File picker is disabled (displaying a list of files without selection is OK)
* [x] Documentation updated
* [x] Manual test case provided

ThomasThelen:
This was completed by bundling the workspace in the zipped Tale. During publication, we extract those files and upload them to DataONE. The file picker was also removed. The documentation was updated in [this](https://github.com/whole-tale/wt-design-docs/pull/81) merged PR

Issue #272, opened 2018-09-17
Improve the file picker


abrinckm:

ThomasThelen:
As part of #305, I'll take this!

ThomasThelen:
The fix for this is to replace the picker with ember's jsTree. The tree takes structured JSON, which means that we need to map each folder/file to a particular format.

I initially did this in the UI, but ran into performance issues so I created an endpoint that returns the structure instead.

ThomasThelen:
Closing this since we've since removed the file picker

PR #430, opened 2019-03-01
Js tree integration

ThomasThelen:
This PR is for replacing the file picker with jsTree. The node list is obtained by querying girder, and is then set as the tree root. 
Fix for #272
Related to https://github.com/whole-tale/girder_wholetale/pull/249

Testing Steps:
This is hard to test without integrating with publishing, which will be covered in issue #305. In the spirit of incremental changes, I'd like to get this piece in before refactoring the modal.

Note: Mike, the last tike you saw this the logic for creating the node tree was in the component, but I've since moved it to girder_wholetale.


Xarthisius:
Per discussion on dev meeting on 05/13/19 this can be closed.

Issue #431, opened 2019-03-04
Add a Feature that Allows Users to Add Multiple Authors to a Tale


ThomasThelen:
Users should be able to attribute multiple authors to their Tale, and we have multiple use cases that support this. This has been implemented in the backend by modifying the authors field as

```
tale['authors'] = [ 
    {
        "firstName": "Kacper",
        "lastName": "Kowalik",
        "orcid": "0000-0003-1709-3744"
    }, 
    {
        "firstName": "Tommy",
        "lastName": "Thelen",
        "orcid": "0000-0003-1709-3754"
    }
]
```

The UI changes to accommodate for this can be split into two parts

### Helper for Parsing tale['authors']
The Tale Cards in the browse page show first and last names of the Tale authors. This was originally using a string value in `tale['authors']`. Since this structure has changed into an array of objects (see above), we need to make a change to parse the new `tale['authors']` and return a string of authors for the cards. 

If this change is not done, the cards will resemble the following (note the authors)
<img width="381" alt="Screen Shot 2019-03-19 at 5 38 45 PM" src="https://user-images.githubusercontent.com/9289652/54650982-060ec500-4a6e-11e9-8dd3-f7328538075c.png">

### Component for Adding Tale Authors
We want users to be able to add custom authors to their Tale. To do this, they'll need to provide a first, last, and ORCID. Following suit of other author-adding systems (see zenodo), we will dynamically create author fields as the user adds them, and remove them when requested.

The natural place for this to occur is in the Tale metadata view under run.

This can be done using [semantic-ui forms](https://semantic-ui.com/collections/form.html#inline-fields).

ThomasThelen:
These features were completed in PR #445

PR #432, opened 2019-03-04
Allow users to edit the Tale Authors field

ThomasThelen:
This is the PR for issue #431. It allows users to modify the Tale Authors field in the metadata view.

To Test:
   1. Launch a Tale
   2. Navigate to the Tale metadata view
   3. Change the Tale authors
   4. Request the Tale from girder
   5. Note that the Tale Authors field has been modified accordingly


craig-willis:
This does what we need for the short term.  I can modify the authors and it looks fine on the browse card.

screen shot 2019-03-05 at 6 09 17 am

PR #434, opened 2019-03-05
Add Functionality to Download a Zipped Tale

ThomasThelen:
### Download Zip
I added a menu item to the three dots dropdown in the Run page. The code for downloading the zip was grabbed from Stack Overflow after failed attempts at doing it my way. Zipping can take some time, and it felt weird not getting any feedback so I put in a semantic ui dimmer that is present until the zip is ready for download.

To Test
   1. Check this branch out
   2. Check out the tale_export branch in girder_wholetale
   3. Deploy
   4. Visit the Run page
   5. Attempt to zip a tale by clicking the zip button in the dropdown
   6. See the dimmer
   7. Get prompted to download the zip (may take a while)

One annoyance I saw is that the dropdown isn't un-selected after selecting an item. I check with semantic-ui and their examples show the same behavior.

### DIFF
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -167,16 +167,10 @@ export default Component.extend(FullScreenMixin, {
             return true;
         },
         
-        exportZip(id) {
+        exportTale(id, format) {
           const token = this.get('tokenHandler').getWholeTaleAuthToken();
           let url = `${config.apiUrl}/tale/${id}/export`;
-          window.open(url + '?token=' + token + '&taleFormat=native', '_newtab');
-        },
-
-        exportBag(id) {
-          const token = this.get('tokenHandler').getWholeTaleAuthToken();
-          let url = `${config.apiUrl}/tale/${id}/export`;
-          window.open(url + '?token=' + token + '&taleFormat=bagit', '_newtab');
+          window.location.assign(url + '?token=' + token + '&taleFormat=' + format);
         },
 
         actionMenuClicked() {
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -167,62 +167,22 @@ export default Component.extend(FullScreenMixin, {
             return true;
         },
         
-        exportTale(id) {
-          let self = this;
-          // Asks the girder endpoint for a zipped Tale
-          // This code was adapted from
-          // https://stackoverflow.com/questions/16086162/handle-file-download-from-ajax-post
-          const token = self.get('tokenHandler').getWholeTaleAuthToken();
+        exportZip(id) {
+          const token = this.get('tokenHandler').getWholeTaleAuthToken();
           let url = `${config.apiUrl}/tale/${id}/export`;
-          var xhr = new XMLHttpRequest();
-          xhr.open('GET', url, true);
-          xhr.responseType = 'arraybuffer';
-          xhr.setRequestHeader("Girder-Token", token);
-          self.set('showDimmer', true);
-          xhr.onload = function () {
-              if (this.status === 200) {
-                self.set('showDimmer', false);
-                  var filename = "";
-                  var disposition = xhr.getResponseHeader('Content-Disposition');
-                  if (disposition && disposition.indexOf('attachment') !== -1) {
-                      var filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
-                      var matches = filenameRegex.exec(disposition);
-                      if (matches != null && matches[1]) filename = matches[1].replace(/['"]/g, '');
-                  }
-                  var type = xhr.getResponseHeader('Content-Type');
-    
-                  var blob = typeof File === 'function'
-                      ? new File([this.response], filename, { type: type })
-                      : new Blob([this.response], { type: type });
-                  if (typeof window.navigator.msSaveBlob !== 'undefined') {
-                      // IE workaround for "HTML7007: One or more blob URLs were revoked by closing the blob for which they were created. These URLs will no longer resolve as the data backing the URL has been freed."
-                      window.navigator.msSaveBlob(blob, filename);
-                  } else {
-                      var URL = window.URL || window.webkitURL;
-                      var downloadUrl = URL.createObjectURL(blob);
-    
-                      if (filename) {
-                          // use HTML5 a[download] attribute to specify filename
-                          var a = document.createElement("a");
-                          // safari doesn't support this yet
-                          if (typeof a.download === 'undefined') {
-                              window.location = downloadUrl;
-                          } else {
-                              a.href = downloadUrl;
-                              a.download = filename;
-                              document.body.appendChild(a);
-                              a.click();
-                          }
-                      } else {
-                          window.location = downloadUrl;
-                      }
-    
-                      setTimeout(function () { URL.revokeObjectURL(downloadUrl); }, 100); // cleanup
-                  }
-              }
-          };
-          xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
-          xhr.send();
-        }
+          window.open(url + '?token=' + token + '&taleFormat=native', '_newtab');
+        },
+
+        exportBag(id) {
+          const token = this.get('tokenHandler').getWholeTaleAuthToken();
+          let url = `${config.apiUrl}/tale/${id}/export`;
+          window.open(url + '?token=' + token + '&taleFormat=bagit', '_newtab');
+        },
+
+        actionMenuClicked() {
+          // We need to initialize the export dropdown each time the 
+          // tale action menu is opened
+          $('.ui.left.pointing.dropdown.link.item').dropdown();
+        },
     }
 });
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -8,7 +8,6 @@ import { A } from '@ember/array';
 import { not } from '@ember/object/computed';
 import $ from 'jquery';
 import layout from './template';
-// import hasEmberVersion from 'ember-test-helpers/has-ember-version';
 
 const O = Object.create.bind(Object);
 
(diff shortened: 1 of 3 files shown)


ThomasThelen:
I agree with the dimmer; I think it looks out of place. One other idea is to use the notification stream so that when clicked, a notification box pops up at the top right notifying the Tale is being zipped.

I'l go ahead and split this into two PRS

ThomasThelen:
A change to the model file made it into the export commit so I tossed out the original two commits and started fresh (force-pushed the tale_export_license_edit  branch from 0422f4a to 114fccf).

The only commit in this branch contains the zip UI integration.

The second force push (force-pushed the tale_export_license_edit  branch from 33b808a to 544ff69 a minute ago) was a rebase with master.

craig-willis:
I'd like @bodom0015 to approve this before merging.

Xarthisius:
That `exportTale` method looks complicated (at least for my puny js skills). What don't we just download the file directly via `window.assign.url`? Something like:

```
exportTale(id) {
  var token = document.cookie.replace('girderToken', 'token');
  let url = `${config.apiUrl}/tale/${id}/export`;
  window.location.assign(url + '?' + token);
}
```
pardon my js. I'm sure there are better ways of getting that token from cookie and constructing final url.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,63 @@ export default Component.extend(FullScreenMixin, {
...
+                      if (matches != null && matches[1]) filename = matches[1].replace(/['"]/g, '');
+                  }
+                  var type = xhr.getResponseHeader('Content-Type');
+    
+                  var blob = typeof File === 'function'
+                      ? new File([this.response], filename, { type: type })
+                      : new Blob([this.response], { type: type });
+                  if (typeof window.navigator.msSaveBlob !== 'undefined') {
+                      // IE workaround for "HTML7007: One or more blob URLs were revoked by closing the blob for which they were created. These URLs will no longer resolve as the data backing the URL has been freed."
+                      window.navigator.msSaveBlob(blob, filename);
```
bodom0015: No need for IE workarounds anymore imo, since Microsoft doesn't even consider it a browser.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,63 @@ export default Component.extend(FullScreenMixin, {
...
+                  } else {
+                      var URL = window.URL || window.webkitURL;
+                      var downloadUrl = URL.createObjectURL(blob);
+    
+                      if (filename) {
+                          // use HTML5 a[download] attribute to specify filename
+                          var a = document.createElement("a");
```
bodom0015: This is a red flag - if you find yourself calling `document.createElement` in a modern JS framework, then there's almost certainly a cleaner approach that will work within the patterns of the framework.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,63 @@ export default Component.extend(FullScreenMixin, {
...
+                          // safari doesn't support this yet
+                          if (typeof a.download === 'undefined') {
+                              window.location = downloadUrl;
+                          } else {
+                              a.href = downloadUrl;
+                              a.download = filename;
+                              document.body.appendChild(a);
+                              a.click();
```
bodom0015: You are adding a download link to the page, and then clicking on the link?

This seems overly complex, and we just fall back to changing the URL anyways when certain features aren't supported

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,63 @@ export default Component.extend(FullScreenMixin, {
...
+          var xhr = new XMLHttpRequest();
+          xhr.open('GET', url, true);
+          xhr.responseType = 'arraybuffer';
+          xhr.setRequestHeader("Girder-Token", token);
+          self.set('showDimmer', true);
+          xhr.onload = function () {
+              if (this.status === 200) {
+                self.set('showDimmer', false);
+                  var filename = "";
+                  var disposition = xhr.getResponseHeader('Content-Disposition');
```
bodom0015: Chrome gives me a (seemingly innocuous?) browser error here:
```
Refused to get unsafe header "Content-Disposition"
```

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,63 @@ export default Component.extend(FullScreenMixin, {
...
+                  if (disposition && disposition.indexOf('attachment') !== -1) {
+                      var filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
+                      var matches = filenameRegex.exec(disposition);
+                      if (matches != null && matches[1]) filename = matches[1].replace(/['"]/g, '');
```
bodom0015: I don't think we need this filename handling..

Since we control the name of the file that will be downloaded, which is currently always a Tale ID, we won't ever have quotes to strip out

craig-willis:
Per 3/21 discussion, use @xarthisius's approach.

ThomasThelen:
I went ahead and made Kacper's recommended change with the caveat that I used window.open so that it could be done in another tab.

Demo http://recordit.co/SwUMLu2uLH

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,23 @@ export default Component.extend(FullScreenMixin, {
         denyDelete() {
             return true;
         },
+        
+        exportZip(id) {
+          const token = this.get('tokenHandler').getWholeTaleAuthToken();
+          let url = `${config.apiUrl}/tale/${id}/export`;
+          window.open(url + '?token=' + token + '&taleFormat=native', '_newtab');
```
bodom0015: I don't think `_newtab` is a valid value for the `target` attribute... I think `_blank` should open in a new tab/window (based on the user's browser preferences), if that's the desired behavior

See:
* https://www.w3schools.com/jsref/met_win_open.asp
* https://www.w3schools.com/tags/att_a_target.asp
* https://stackoverflow.com/questions/4964130/target-blank-vs-target-new/8867079#8867079
* https://stackoverflow.com/questions/18360438/attribute-target-newtab-for-links

app/components/dashboard/run/left-panel/template.hbs
```diff
@@ -41,13 +28,22 @@
...
                         {{#click-outside action=(action (mut displayTaleInstanceMenu) false)}}
                             <div class="tale-instance-menu">
                                 <div class="ui vertical left menu transition">
                                     <a class="item" href="https://wholetale.readthedocs.io/en/stable/users_guide/run.html" target="_blank">
                                         Read the docs <div class="ui label transparent"><i class="fas fa-book"></i></div>
                                     </a>
+                                    <div class="ui left pointing dropdown link item">
+                                      Export
+                                      <div class="menu">
+                                        <div class="item" style="width: 200px; z-index:999;" {{action "exportBag" model.taleId}}>BagIt</div>
```
bodom0015: We may need to revisit this `z-index` value as we integrate upcoming new features - I know the `notification-stream` component uses a `z-index` of 1.

As we move forward, we can decide which component(s) takes precedence and increase their `z-index` values to show them on top.

app/components/dashboard/run/left-panel/template.hbs
```diff
@@ -129,4 +104,13 @@
...
     </div>
+
+     {{#if showDimmer}}
+    <div class="ui segment" style="position:fixed; top:25%; width:50%; left:25%; height:35%; z-index:999">
+      <div class="ui active dimmer">
+        <div class="ui indeterminate huge text loader">Zipping your Tale</div>
+    </div>
+    <p></p>
+  </div>
+  {{/if}}
```
bodom0015: I didn't see the dimmer in your new GIF... if it is no longer going to display, we should probably just remove it

Xarthisius:
I have weird issue in Chrome:

ala

dropdown disappears immediately when I click on it. As result I cannot export at all. Works fine in Firefox.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,23 @@ export default Component.extend(FullScreenMixin, {
...
+        },
+
+        exportBag(id) {
```
Xarthisius: Could you collapse that into one function by passing `taleFormat` as 2nd argument?

app/components/dashboard/run/left-panel/component.js
```diff
@@ -165,5 +166,23 @@ export default Component.extend(FullScreenMixin, {
...
+          const token = this.get('tokenHandler').getWholeTaleAuthToken();
+          let url = `${config.apiUrl}/tale/${id}/export`;
+          window.open(url + '?token=' + token + '&taleFormat=bagit', '_newtab');
```
Xarthisius: Why would we want to open a tab if it's immediately closed? It just causes my web browser to flicker. I believe my original suggestion didn't exhibit this behavior.

app/components/dashboard/run/left-panel/template.hbs
Xarthisius: I believe you can pass multiple arguments to an action like this: `{{action "exportTale" model.taleId "native"}}`

Also, can the width of this menu item match the others?

Xarthisius:
Thanks @ThomasThelen !

ThomasThelen:
### Changes Addressing Comments

#### Use Kacper's URL Location
I was testing it on my localhost:4200 and not on my VM so I wasn't getting a redirect back to the home page (I was getting sent to the api URL) so I had replaced that section with window.open(). Testing the window.location method on my VM didn't redirect me to the backend API and works fine, so I used it.

I also combined both export functions into a single, taking the query param as a handlebars template parameter.

#### Remove Dropdown
There were a few things that cropped up styling wise and the issue Kacper mentioned above. 
I believe the issue Kacper ran across is that there is a need to call `.dropdown()` on the dropdown div _before_ it gets clicked. The issue is that the div doesn't exist during the page's onload, or when the parent div is clicked. In Kacper's case it was getting called on a click to the dropdown, which was probably the flickering.

I noticed if there is only one Tale the background of the bagit and zip dropdown items blends in the with background, and setting it a little darker via css messes up the mouse over color change that's built in with the semantic-ui component.

I removed the dropdown until we have time to look into those issues.

#### Remove Dimmer Code
Since we no longer have knowledge when the api request finishes and because the browser's page spinner shows that a request is being processed, we don't need this anymore.


#### Added Icons
I added an icon for each download, seen below
Screen Shot 2019-03-21 at 7 50 35 PM

PR #443, opened 2019-03-14
Notification widget to receive status updates on "Rebuild"

bodom0015:
### Problem
Rebuild needs to notify user when build completes or fails. User needs to be able to view the logs from failed events to fix the broken build.

### Approach
Implement SSE using a custom `EventSource` allowing us to pass the `Girder-Token` header. Also added a log viewer modal that will display pre-formatted text

NOTE: There may still be formatting errors in the log output that we should revisit (e.g. random commas, escape sequences, console codes, etc)

### How to Test
3. Launch a Tale, if you haven't already
4. Navigate to the "Run" view and select a Tale
5. Expand the "..." dropdown at the top-center, and choose "Rebuild Tale"
    * After a few seconds, you should see the new `notification-stream` component appear to tell you that the image is building
6. Click on the status (e.g. `BUILDING`) to see the logs
    * You should see that the log output is automatically updated/refreshed every ~2 seconds
    * Once the build is complete, you should receive another notification indicating that your build is successful (e.g. `SUCCESS`)
    * You should be offered on option to "Restart to Use New Image"
7. Modify the `apt.txt` in your workspace to contain a nonsense package name (e.g. `q; isdhadshiasd `)
8. Select "Rebuild Tale" again
    * After a few seconds, you should see the new `notification-stream` component appear to tell you that the image is building
    * After a few more seconds, you should see a new notification appear indicating that the image build failed (e.g. `INVALID`)
9. Click "Mark All as Read" atop the notification-stream
    * Your notification-stream should dismiss itself
    * Subsequent refreshes should **not** bring back dismissed notifications

NOTE: there is still a subtle timeout issue with EventSource's integration with Girder - see https://github.com/Yaffle/EventSource#server-side-requirements

### DIFF
--- a/app/components/dashboard/unified-view/template.hbs
+++ b/app/components/dashboard/unified-view/template.hbs
@@ -8,3 +8,5 @@
     </div>
   </div>
 </div>
+
+{{ui/notification-stream}}
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -3,6 +3,7 @@ import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import config from '../../../config/environment';
 import layout from './template';
+import JSONBeautify from 'npm:json-beautify';
 
 const EventSource = window.EventSourcePolyfill;
 
@@ -44,14 +45,8 @@ export default Component.extend({
         events.push(event);
         this.set('events', events);
         let json = JSON.parse(event.data)
-        if (json.type == 'job_created' && json.data.title == 'Build Tale Image') {
-            alert("Building Tale Image: " + json.data.args[0]);
-        } else if (json.type == 'job_status') {
-            alert("Tale Building: " + json.data.args[0] + "\nStatus is now " + json.data.status);
-        } else {
-            alert("Unknown job type: " + json);
-            console.log(json);
-        }
+        console.log("JSON data recv'd:", json);
+        alert(event.data);
     },
     onError(error) {
         console.log("EventSource failed.");
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -1,0 +1,69 @@
+// app/pods/components/markdown-component/component.js
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+import config from '../../../config/environment';
+import layout from './template';
+
+const EventSource = window.EventSourcePolyfill;
+
+export default Component.extend({
+    layout,
+    tokenHandler: service('token-handler'),
+    
+    apiHost: config.apiHost,
+    events: [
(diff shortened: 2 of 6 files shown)


craig-willis:
Overall this looks great.  I like the design and, aside from a few problems below, it's very usable.

I'm seeing more notifications than expected -- what I think may be duplicates -- and they remain even after hitting Mark All as Read. At first I thought this was something on the server end, but since I'm expiring notifications on the backend, I now think it's a Dashboard thing. ~10 minutes after a single build, I now have 20 entries alternating between "building" and "success" whereas the backend notification stream is empty (i.e., everything expired)

Screen Shot 2019-03-20 at 9 24 30 PM

```
$ curl -X GET --header 'Accept: application/json' --header 'Girder-Token: xxx' 'https://girder.local.wholetale.org/api/v1/notification'
[]
```

I also noticed that the log entries for each are identical. This seems similar to the problem we had earlier with the page itself appending duplicates over time?

On the timeout issue, have you seen the [discussion of the need for a heartbeat/comment messages](https://github.com/Yaffle/EventSource/issues/24) and it looks like this is now [maybe configurable](https://github.com/Yaffle/EventSource/blob/767dc9c621af9e43540b4556d6dcdbc502d8af5b/src/eventsource.js#L638) and has a 45 second default.

A few other minor points:
* Position of notification box is a little inconvenient because it blocks interaction with the launched Tales. Maybe position it at the bottom right?
* Wrench icon does not indicate job or notification to me.
* It might be good to have explicit "view logs" link instead of clicking on the build/success status
* Tale name link doesn't seem to do anything

bodom0015:
To address @craig-willis' other points:

> Position of notification box is a little inconvenient because it blocks interaction with the launched Tales. Maybe position it at the bottom right?
It should be easy to move it to a new static location, although I'm unsure how easy it would be to fix it to the bottom-right. If that proves too difficult, we could always bring back a navbar button similar to the one that previously toggled the `job-watcher`

> Wrench icon does not indicate job or notification to me.
To me, wrench was the best thing I could find that could indicate "build" ... I'm open to suggestions, if you have a better one in mind. I wanted to use "hammer" or "rev" or "tools", but we are only on version 5.0.8 of font-awesome, and many of the nicer icons come in later versions. For some reason, the font-awesome dependency is not pulled in via `node_modules`, so it was unclear if there were any customizations that would be removed if I were to update to a newer version.

> It might be good to have explicit "view logs" link instead of clicking on the build/success status
I agree, but was worried about each notification feeling too cluttered. I can work on improving the design of the build events while we work on expanding the types of events that the UI receives and handles

> Tale name link doesn't seem to do anything
Once we finish the refactoring from `/run/:instance_id` to `/run/:tale_id`, my plan was to link to the newly-built Tale's run view here... until that is ready I can either link to the Tale View page or remove the link entirely

PR #448, opened 2019-03-26
Copy on Launch

bodom0015:
### Problem
When a user launches a Tale that they do not have write access to, they are left with a half-working Tale on which they can't actually add any data or create any files. 

### Approach
When attempting to launch an instance that the user cannot access (e.g. `_accessLevel < 1`), prompt the user with a modal to copy the Tale. If the user confirms, call the new `/tale/:id/copy` endpoint before launching. This will create a copy of the existing Tale that the current user can write to, allowing them to rerun and expand the Tale of another user. We then launch this copy using the existing methods.

#### TODOs (subsequent issues)
* Don't allow user to make multiple copies
* Tie in with event notifications via `notification-stream`
* (Possibly) get rid of the modal altogether and continue without user confirmation

### How to Test
Prerequisite: at least one Tale that you do not have access to

1. Checkout and run this branch alongside [girder_wholetale's 232_tale_copy branch](https://github.com/whole-tale/girder_wholetale/pull/233), rebuild the dashboard
3. Navigate to Browse
4. Locate a Tale that you cannot access
    * NOTE: To easily locate an inaccessible Tale, they will be missing their "delete" button
5. Click "Launch Tale" on the inaccessible Tale
    * A modal should appear prompting you to copy the Tale
6. Confirm the copy operation
    * You should see the launching Tale appear on the right side
    * You should see a status update appear beneath the launched Tale on the left side
7. Click the Tale Instance on the right side
    * You should be brought to the Run view with the new Tale instance selected
8. Navigate to Run > Files, select Tale Workspace, and Upload a File or Create a Folder
9. Once the container instance finishes starting, navigate to Run > Interact
10. Open a terminal in Jupyter or RStudio (whichever type of Tale you launched) and perform an `ls -al`
    * You should see the files/folders you uploaded in step 8 above - this confirms that you have Write Access to the new Tale Copy


abrinckm:
Everything works very well up until steps 8 and 10.

In step 8, when I create a new folder inside "workspace" it disappears when I refresh the screen and reappears in "home". Sounds like an issue unrelated to this PR.

In step 10, I open up a terminal and do the following: (here is what I see):
```
jovyan@b8b65df0d930:~/work/workspace$ ls -la
total 8
drwxr-xr-x 2 root root 4096 Jun 11 15:19 .
drwxr-xr-x 1 root root 4096 Jun 11 15:19 ..
jovyan@b8b65df0d930:~/work/workspace$ mkdir test_create_folder
mkdir: cannot create directory ‘test_create_folder’: Permission denied
jovyan@b8b65df0d930:~/work/workspace$ sudo mkdir test_create_folder
bash: sudo: command not found
jovyan@b8b65df0d930:~/work/workspace$
```

So it doesn't look like I have write permissions. 

Other than that it worked as expected.

bodom0015:
Thank you for the feedback, @abrinckm! I can run through the test case again as well to reproduce this and dive into what might be happening there.

@Xarthisius could this possibly be an issue with the related [PR for girder_wholetale](https://github.com/whole-tale/girder_wholetale/pull/233)?

craig-willis:
@abrinckm I just ran through a test of the PR and may have an explanation for what you saw in step 8 but did not have problems with step 10.  

Prior to testing, I did a `make clean && make dev` using the latest `deploy-dev:master` branch.  However, I did not clear local storage in the browser for dashboard.local.wholetale.org.  When I selected the "Tale Workspace" folder via the UI, it was empty and console reported 400 errors on the `workspaceRootId`.  The value in local storage was from before I cleaned my local database and therefore invalid.    Can you retry first without clearing local storage to see if that's the case (i.e., 400 errors when loading the workspace)?

I was able to create folders from both the Jupyter environment and WT UI and ran the notebook without problem.

I'd like to get Kristina's input on the "Insufficient Access" window and message. I think it might confuse users and could just be something informational ("Hey, since you don't own this we have to make a copy so you can edit/run it. It will show up under 'your tales'. Got it?")

Otherwise, looks good.

abrinckm:
@craig-willis If it's cache then I'll try again running the dashboard incognito.
one important difference to note: I had changed the girder branch to 232_tale_copy_branch before running make clean & dev. I will try again just running master as you did.

abrinckm:
@craig-willis @bodom0015 

Just updating so everyone is aware. The problem was on my end. I am getting this error:
```
[2019-06-14 14:48:28,432: INFO/ForkPoolWorker-2] Calling: girderfs -c wt_work --api-url http://10.0.0.2:8080/api/v1 --api-key QY08mjfIUlfZ39RaQjFq35yYU4I2SBDxjxg7Ukpy /var/lib/docker/volumes/5d03b3b23bb19857b7022d06_abrinckm_1O2U9C/_data/workspace 5d03b3b23bb19857b7022d06
open: Permission denied
mount.davfs: can't mount using fuse kernel file system
mount.davfs: trying coda kernel file system
mount.davfs: no free coda device to mount
```
Craig is looking for a solution

Xarthisius:
Works great, thank @bodom0015 !

Issue #281, opened 2018-10-22
Cut Stack Trace out of Publishing Errors


ThomasThelen:
The whole stack trace is getting displayed in the UI when an error happens. Before the stack trace, a custom message ("You need to sign into DataONE", "This Tale is too old to publish", "Could not retrieve file ...") is shown.

We should only show the custom message and discard the stack trace.

ThomasThelen:
This can be done by querying the new result endpoint in the job endoint

ThomasThelen:
This was completed in the publishing branch, [here](https://github.com/whole-tale/dashboard/commit/16ba139656b30f994fc74ba4700c6b5783431164)

PR #451, opened 2019-03-29
Publish modal styling changes

### LINKED ISSUE #446: Multiple publishing problems
Noting multiple problems found during review of https://github.com/whole-tale/dashboard/pull/344 that we will address after the initial PR is merged.

* Modal should not extend to the bottom of the page 
* Styling should be consistent with other modals (e.g., register)
* The "accordion" layout was part of the [original mockup](https://projects.invisionapp.com/freehand/document/nPpPcHSRg) but seems unnecessary and inconsistent with the rest of the site
* Modal should disable "Publish" button and change "Cancel" to "Close" on success.
* REST endpoint errors should be reported in the UI (e.g., 403 if publishing a Tale you don't own or 500) via ember.onerror
* Job errors should propagate via notification stream.
* We may consider re-adding the informative list of files to be published, but not as a selectable tree.

bodom0015:
### Problem
The Publish Modal is inconsistent with the styling elsewhere in the application.

Fixes #446

### Approach
Refactor look and feel of the Publish Modal slightly to better match other parts of the WholeTale Dashboard. This also includes some refactoring / better error-handling for the `notification-stream`, as we plan to soon enhance the publishing task to provide asynchronous updates via Girder's Notification API.

### How to Test
2. Open a new incognito window and login to the WholeTale Dashboard
3. Compose and Launch a Tale, if you haven't already - ensure that this is a Tale that you own
    * NOTE: Once #448 is merged, you should no longer be given an option to Launch tales that you do not own
4. Navigate to Run and select a Tale
5. Click the ellipses on the `wt-paddleboard` to open a dropdown, then click "Publish Tale..."
    * You should see a modal appear, prompting / redirecting you to go to the DataONE login page
    * Once logged in, you should be redirected back to the WholeTale Dashboard (with `?auth=true` appended to the URL)
6. Navigate once again to Run, select a Tale, expand the dropdown, and click "Publish Tale..."
    * This time, the Publish Modal should appear in the view
    * The "Target Repository" dropdown should automatically expand for you, saving you a mouse click
7. Select "DataONE Development" from the dropdown and click "Publish"
    * You should see the "Publish" button change to "Publishing", and its checkmark should change to a loading symbol - the button should be disabled at this time
    * You should see a progress bar appear in a blue message at the bottom of the modal
8. Wait for the publishing operation to complete successfully
    * You should see the "Publishing" button change to "Published", and its checkmark should reappear - the button should still be disabled, preventing a 2nd publish without reopening the modal
    * You should see the blue progress message has been replaced with a green success message

NOTE: In the case of error, the green success message is replaced with a red error message. This error message should display any error that comes back from the publishing API or Job.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -19,7 +19,11 @@ const App = Application.extend({
 loadInitializers(App, config.modulePrefix);
 
 Ember.onerror = (error) => {
-  console.log("Ember.js encountered an error:", error);
+  if (console.error) {
+    console.error(error);
+  } else {
+    console.log("Ember.js encountered an error:", error);
+  }
 };
 
 console.log("The app is running in " + config.environment + " mode");
--- a/app/app.js
+++ b/app/app.js
@@ -18,6 +18,10 @@ const App = Application.extend({
 
 loadInitializers(App, config.modulePrefix);
 
+Ember.onerror = (error) => {
+  console.log("Ember.js encountered an error:", error);
+};
+
 console.log("The app is running in " + config.environment + " mode");
 console.log("The API is " + config.apiUrl);
 export default App;
--- a/app/components/dashboard/compose/left-panel/component.js
+++ b/app/components/dashboard/compose/left-panel/component.js
@@ -289,7 +289,8 @@ export default Component.extend({
               // The job is running or queued
               if (job.progress) {
                   // Update the component variables related to progress
-                  component.set('progress', job.progress.current/job.progress.total);
+                  component.set('progress', job.progress.current);
+                  component.set('progressTotal', job.progress.total);
                   component.set('message', job.progress.message);
               }
             }
--- a/app/components/dashboard/compose/left-panel/template.hbs
+++ b/app/components/dashboard/compose/left-panel/template.hbs
@@ -17,11 +17,13 @@
     {{#if launchingInstance}}
       <div class="compose-status">
           <p class ="compose-status-message">
-            {{message}}
+            <i class="fas fa-spinner fa-pulse"></i> {{message}}
           </p>
         {{#if importing}}
           <div class ='tale-compose-progressbar'>
-            {{ember-progress-bar progress=progress options=(hash strokeWidth=4 color=barColor trailColor= "#a5a4a4")}}
+            {{#ui-progress value=progress progress=progress total=progressTotal class="teal indicating"}}
+              <div class="bar"></div>
+            {{/ui-progress}}
           </div>
         {{else}}
           <div class="tale-compose-loader">
--- a/app/components/dashboard/manage/left-panel/component.js
+++ b/app/components/dashboard/manage/left-panel/component.js
@@ -1,8 +1,9 @@
 import TextField from '@ember/component/text-field';
+import { A } from '@ember/array';
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import { observer, computed } from '@ember/object';
-import Object from '@ember/object';
+import EmberObject from '@ember/object';
 import $ from 'jquery';
 
 function wrapFolder(folderID, folderName) {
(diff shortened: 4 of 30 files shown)


craig-willis:
This looks great, @bodom0015. It addresses most of the issues identified in https://github.com/whole-tale/dashboard/issues/446. 

I think we should have explicit handling for 403 errors with a custom message.  Currently, I'm seeing:

Screen Shot 2019-03-30 at 4 38 12 AM

Something like "You must be the Tale owner to publish" would be clearer.

For all Girder errors, there's usually a message on the response that can be displayed (maybe behind a "details" link, since they are not particularly friendly:

```
{"message": "Admin access denied for tale 5c9f38280bb05d0001559eb4 (user 59fb6165f7e8790001da4e8b).", "type": "access"}
```

ThomasThelen:
Wow this looks pretty awesome; nice job!

I think we're using the job/result endpoint which is supposed to give the error message instead of a full trace (this was issue #281).

Craig's error might have occurred before the job was created, which is why we're not seeing a message on the UI.

craig-willis:
@ThomasThelen with https://github.com/whole-tale/girder_wholetale/pull/274/commits/60acf10415fc63457a7f2eddc357cb04db74f118 we modified the endpoint to use Girder's built-in permissions to handle 403 instead of checking in the job.  As a side note, the Dashboard was ignoring other errors returned by the endpoint (e.g., 500). So we should be handling both with this PR -- errors returned from the endpoint and errors reported in the job itself.

bodom0015:
Great catch, @craig-willis... I had tested failure in the job by handing a bogus jwt to the `publishTale` call. @ThomasThelen was right in this case, as it seems I had forgotten to set the `statusMessage` if a REST encountered while submitting the job. Thanks for catching that!

The error / success message should actually already be coming from the `getFinalJobStatus` call that @ThomasThelen mentioned.

I also went ahead and hid the menu option if you aren't an Admin of the Tale, to prevent future confusion surrounding that particular failure. I'm sure there's plenty of error edge cases there, and some that I've likely missed so if anybody can think of other clever ways it could break, please bring them up here :)

bodom0015:
I should also note that one feature has not been preserved: there was quite a bit of code to check if an existing publish job was running/had already run. My reasoning was that this seems like an unnecessary burden on the frontend, as the backend could more easily enforce such things.

For example, an attempt to submit a second job simultaneously to the API could cause the API to simply return the currently running job instead. If the API receives a new request to publish a Tale that hasn't changed since the last time it was successfully published, then the API could simply return 200 (or maybe 400, depending how we'd like to frame this to the user).

While I haven't made these changes yet, they don't seem overly complicated (individually)

bodom0015:
I attempted to go back and cleanup the (now duplicated) publishing logic and refactor this into a component as @ThomasThelen had done. Along the way, I seem to have somehow broken something and now get `TypeError: Cannot read property 'get' of undefined` looping in the console, coming from some internal library in `vendor.js` that I don't recognize.

In order to avoid polluting this working branch with untested/unstable changes, I pushed the WIP cleanup stuff to a separate branch called [publish-modal-cleanup](https://github.com/whole-tale/dashboard/tree/publish-modal-cleanup). If I have time to continue refactoring this later on, I will try to pick up where I left off.

In the meantime, I think this should be ready for review.

bodom0015:
Great catch, @ThomasThelen! I've updated `publishedURL` in the `tale-tab-metadata` component to be a computed value. This way when the `tale` or `tale.publishInfo` changes underneath it, Ember knows to re-evaluate this value to the latest publishInfo entry's `uri`.

While testing this, I noticed an issue integrating authors and publishing that may require slightly better error-handling: When you Compose a new Tale we don't know your ORCID (a required field), so the author information can't be filled out properly and is created as invalid in Girder. We can't leave the authors empty either, as this is also invalid.

Attempting to publish an invalid Tale (for me) yielded a cryptic error message stating something along the lines of `__init__() missing 3 required positional arguments: 'text', 'url', and 'method'`

Issue #213, opened 2018-07-27
Deleted instances still show up in Launched Tales tab


Xarthisius:
Steps to reproduce:

1. Launch a few Tales (4 should be enough)
2. Delete them one by one from the right-hand side menu
3. Change tab

Result:
Usually at least one of the instances will pop back up in the menu. Our working theory is that, widget doesn't wait for 200 from DELETE, and `GET /instance` called upon switching view is still able to pick it up.

Expected result:
Tales are removed from the panel, when they're actually gone from the backend, not by just modifying ember state.

swyngaard:
This error is related to the error I mentioned in #185 . I believe the error stems from transitioning routes after deleting an instance. I still need to report the bug to the Ember devs.

craig-willis:
As discussed with @cyberdante on dev call, is there a workaround in the meantime?

cyberdante:
This is unfortunately still outstanding.

bodom0015:
As @ThomasThelen suggested, this appears to be a side-effect of the asynchronous nature of deleting a container instance. One way to help this issue from our end might be to add a new DELETING state to [`InstanceStatus`](https://github.com/whole-tale/girder_wholetale/blob/master/server/constants.py#L37-L50).

This way, when the DELETE `/instance/:id` endpoint is hit we can instantly mark the instance as `DELETING`, signaling that the instance will soon be deleted. We can then add logic to the UI that can filter out any instances in this state. Girder/Docker can take as long as they need to clean things up, and the user will no longer be offered that instance in the UI as if it were still valid.

Would this sound like an agreeable solution, if I can get a working implementation?

bodom0015:
I have so far been unable to reproduce this reliably, but found an edge case that may possibly be related to this bug. I figured I should post it here for good measure.

In my case:
1. On the Tale Tab Selector, I used the Tale Metadata Tab to edit the name of a tale and Save
2. After editing the name, I launched a new instance from it immediately (e.g. did not refresh)
    * BUG: the instance was given the old (incorrect) name, instead of the new one
3. Killed the instance I just launched
4. Refreshed until the instance was gone
5. Launched another instance from the renamed tale
    * Instance now had the new (correct) name

Repeating these steps, I am unable to get an instance to launch with the stale/old/incorrect name, so it's unclear if this is directly related, but I suspect this deletion behavior may have played a role.

ThomasThelen:
[...]

PR #459, opened 2019-04-11
Refactor Run View to use taleId instead of instanceId

bodom0015:
### Problem
We don't currently have an integrated view of a Tale's files and metadata without running an Instance of the Tale. This is unnecessary, since the filesystem and metadata exist independently of the running container, and could be operated on separately.

### Approach
The `/run/:id` view currently takes an optional instance ID, which assumes that we have a running container. If this view instead were to take a Tale ID, we can perform a lookup to see if running Instance exists for the Tale.

This presents a few new cases for the Run view, detailed below.

#### Running Tale (Instance Exists, and is `Running`)
* The current run view, more or less
* Interact tab should display an iframe or button to open in new tab
* Additional "Stop" button added at the top-right

#### Stopped Tale (No Instance Exists)
* The current run view, more or less
* Interact tab should display a placeholder message
* Additional "Start" button added at the top-right

#### Transitioning Tale (Instance Exists, but status is not `Running`)
* The current run view, more or less
* Interact tab should display a placeholder message and loading indicator
* Disabled "Please wait" button added at the top-right - this button re-enables itself ~30s after submitting a start or stop request to allow the user to attempt to submit a stop operation.

### How to Test
2. Close all incognito windows and, using a fresh one, login to the WholeTale Dashboard
    * You should immediately notice that the right-hand panel for "Launched Tales" is now gone
3. Compose a Tale if you haven't already
4. From the `Browse` page, click "View" for a Tale
    * You should no longer instead be brought to "Tale View" page when clicking this link
    * You should instead be brought to `Run` view for the Tale
5. Make sure the Tale is stopped - click `Stop` at the top-right, if necessary
    * With a stopped Tale, you should see a placeholder message on Interact, but still be able to use the Files/Metadata tabs
6. Click `Start` at the top-right to Start this Tale
    * You should see the "Start" button change to "Please Wait" with a loading indicator
    * You should see the placeholder message change to "Please Wait" with a loading indicator
    * You should see event notifications coming through, indicating that your Tale image is being rebuilt
7. Select the `Interact` tab and wait for the Tale to finish starting
    * Once the Tale is started, you should see the "Please Wait" on the button at the top-right has changed to "Stop"
    * If the chosen Tale uses an iframe, you should see the placeholder message has been replaced with an iframe to the Tale
    * If the chosen Tale doesn't use an iframe, you should see a placeholder message indicating that the user needs to access the Tale using a new browser tab
    * You should still be able to use the Files / Metadata tabs as before
8. Click `Stop` at the top-right
    * You should see the original placeholder message return
    * You should still be able to use the Files / Metadata tabs as before

#### TODOs
* The "Compose" flow no longer needs to start the Tale for the user upon creation - it is unclear whether the UI or the API currently handles this
* We should now be free to clean-up massive amounts of code from the UI (e.g. defunct controllers, old components, unreachable routes/views, etc)


ThomasThelen:
This is a bit bigger than I thought it would be; I'm going to take a closer look at this when 0.7 is out.

From skimming over in Firefox, 

1. I think we should change `Launch this Tale to Interact` to `Run this Tale to Interact`, or change the `Run` button to `Launch`
2. Launching Tales on the browse page doesn't seem to have any effect (I get `Launching Tale ... with the spinner).

The run page state looks pretty good; I navigated away and back while launching and stopping the Tale and things seemed to look okay.

bodom0015:
Thanks for taking a look at this, @ThomasThelen! 

> I think we should change `Launch this Tale to Interact` to `Run this Tale to Interact`, or change the Run button to Launch

I'm happy to change the wording if we can come to consensus on what it should say. My understanding is that the `Run` nav on the top navbar is eventually going away (per Kristina's mockups), so `Run` may not be the best choice either. Since the Browse and Run views both say "Start", maybe this should instead be `Start this Tale to Interact`? We could also offer an additional button there to `Start the Tale`. 

These might be great topics to bring up tomorrow at the UI/UX meeting.

> Launching Tales on the browse page doesn't seem to have any effect (I get `Launching Tale ... with the spinner).

I am seeing this now as well.. perhaps I botched fixing one of the merge conflicts there, but will take a closer look.

bodom0015:
Updated this PR to fix error-handling for launching a Tale from the "Browse" and "Run" - I had apparently added a catch to the wrong Promise in both cases :X 

Thank you for pointing out such a silly oversight!

ThomasThelen:
Can't wait to see this in production!

The [Return to Dashboard](https://github.com/whole-tale/dashboard/pull/459/files#diff-51cf2528add713a8ed071d3b0b504de0R2) link felt a little weird since it brings you to the browse page (and we're already on the dashboard). I would vote to remove this.
Screen Shot 2019-06-04 at 11 21 16 AM


I'm still running across the infinite spinner on the browse page when clicking `Launch`. I do, however, see the notifications. I checked the console and I didn't see any errors.

bodom0015:
@ThomasThelen thanks for taking another look at this! :)

The "Return to Dashboard" link was taken directly from the mockups, but we can possibly discuss alternatives this afternoon at the UI/UX meeting.

As for the infinite spinner issue, I can take a closer look and try to fix the browse view there to avoid breaking the view completely in between merging these refactoring branches... if I can't avoid a merge conflict between the two branches here, I may just merge the `refactor-browse-tales` into this one, if that isn't too overwhelming to review?

Thoughts?

bodom0015:
@ThomasThelen the spinners seem to update properly for me in Chrome under ideal circumstances. Maybe I'm just missing an intermediate step when testing? When a new Tale is launched, a build could be happening in the background that can take some time to complete, but it should eventually complete. For example, launching two Tales where one has a long build in the background: http://recordit.co/n4KhNNUmZR

There is an edge case here that is not currently handled by the platform that may perhaps be what you're hitting? I've noticed that if you refresh while a Tale is launching, then the loop that is querying for the instance status doesn't restart after the refresh completes. So if you're on the browse view (for example, waiting for a launch to complete) and you refresh the page, you might notice that your "Networks" tab in the developer console is no longer firing `GET /instance/:id` every couple seconds like it should. I'm exploring other edge cases here, such as navigating to different views while launching as well to see if the same behavior appears.

Does this sound consistent with what you're seeing? Is it just the spinner at the right on the (`tale-instances` component) that infinitely spins, or does the spinner under the Launch button on the `tale-browser` also continue to spin?

Xarthisius:
@bodom0015 overall it looks good to me! One thing I'd like to see fixed is the infinite spin on compose if Tale limit is reached:
dash

Though I haven't checked whether it's a regression introduced here or not.

bodom0015:
@Xarthisius this was indeed a regression, good catch! :+1:

I had adjusted `postInstance` to use a Promise and had forgotten to come back and fix it in this view as well.

bodom0015:
@craig-willis feedback noted, and will await further instructions following the UI/UX discussion(s)

> The only possible issue I came across is that if I hit "Stop" then "Return to Dashboard" the instance still showed in the Running Tales panel and clicking on it gave a 404 in the iframe. This happened 2 times out of 5, but I can't repeat reliably

From what I understand, this has always been true when shutting down/deleting an instance (see #213). Refreshing the page quickly after deleting an instance will show the same behavior. And while I agree that that issue is something we should address, I believe it's easier to repeat the 404 problem by following these steps:
1. Launch a Tale
2. Navigate to the Run View
3. Once the instance finishes starting up, click Stop
4. Immediately refresh the page
5. Click Run again

Actual Results: 404 in the Interact window
Expected Results: ??? error message?

I believe that this (like #213) is because the instance still exists in Girder when we ask for it, but no longer points to a running container... how exactly should we handle this? Once again, I propose adding a new `InstanceStatus.DELETING`, as this would be make such a state easier to detect on page load.

craig-willis:
@bodom0015  That last commit fixed what I was seeing. A approve again!

bodom0015:
Sorry, had a set of uncommitted changes that I was unaware of.. I start on the Copy on Launch additions (on another branch) and didn't realize the `taleTransitioning` stuff wasn't in this branch.

Thank you for pointing out that it was missing! :)

Xarthisius:
When I compose a Tale (step 3.) I get error message:
```
Error Creating Tale
Unexpected token < in JSON at position 0
```
which seems bogus, cause the Tale starts without any problems... I don't see anything relevant in console logs.

Xarthisius:
> When I compose a Tale (step 3.) I get error message:
> 
> ```
> Error Creating Tale
> Unexpected token < in JSON at position 0
> ```
> 
> which seems bogus, cause the Tale starts without any problems... I don't see anything relevant in console logs.

After quick debugging it happens here: https://github.com/whole-tale/dashboard/blob/f3dd43051d468b011df20cd2318cb17bd354f9b5/app/components/dashboard/compose/left-panel/component.js#L214

`item` is already and object and subsequent `JSON.parse` yields the error I'm seeing.

bodom0015:
Good catch, hadn't noticed that after fixing the issue you pointed out above in https://github.com/whole-tale/dashboard/pull/459#issuecomment-503172303.

Thanks, @Xarthisius!

PR #461, opened 2019-04-15
Improve notification-stream UX

bodom0015:
### Problem
The notification stream covers up important bits of the UI, with no way to get rid of it unless the events are acknowledged. This presents problems if the user needs to click beneath the feed, but is not yet ready to acknowledge all notifications.

### Approach
* Refactored `notification-stream` to move the actual `events` array out of the component and into the service
* Added a button to the navbar to toggle display of the notification stream
    * Moved existing `notification-stream` component to `header.hbs` / `application.js`
    * Added a badge/label on the button to show the number of notifications currently displayed at a glance

#### TODOs
* Still need to investigate the ability to "update" an existing notification - valid test case needed
* Investigate using a draggable modal for this? Save last coordinates in localStorage? 

### How to Test
3. Compose or Launch a Tale
    * You should see the event feed automatically show itself when a new event arrives (e.g. Tale image is building)
4. Click the new (!) button in the Dashboard's Navbar
    * The event feed should be hidden
    * You should still see the red counter/badge indicating how many events are in the feed
5. Click the (!) button again
    * The event feed should redisplay
6. Click the "Mark All as Read" button in the event feed
    * The event feed should be hidden once again
    * You should no longer see the red counter/badge, as there are no events in the feed
7. Click the (!) button one more time
    * The event feed should redisplay
    * You should see a placeholder message indicating that you have received `No new events`

### DIFF
--- a/app/components/dashboard/unified-view/template.hbs
+++ b/app/components/dashboard/unified-view/template.hbs
@@ -8,5 +8,3 @@
     </div>
   </div>
 </div>
-
-{{ui/notification-stream}}
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -106,6 +106,10 @@ export default Component.extend({
             self.set('selectedEventLogs', []);
             $('#log-viewer-modal').modal('hide');
         },
+        
+        gotoTale(taleId) {
+           this.router.transitionTo('run.view', taleId); 
+        },
     }
 });
  
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -2,21 +2,21 @@
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import { A } from '@ember/array';
-import EmberObject from '@ember/object';
+import EmberObject, { computed } from '@ember/object';
 import config from '../../../config/environment';
 import $ from 'jquery';
 import layout from './template';
 
 export default Component.extend({
     layout,
-    notificationStream: service('notification-stream'),
     
+    notificationStream: service('notification-stream'),
     apiCall: service('api-call'),
     store: service(),
-    logInterval: null,
-    isDev: config.dev,
     
-    events: A([]),
+    logInterval: null,
+    DEBUG: config.dev,
+    
     source: null,
     selectedEvent: null,
     selectedEventLogs: [],
@@ -25,15 +25,7 @@ export default Component.extend({
         const self = this;
         const source = self.get('notificationStream').connect();
         
-        source.onopen = () => console.log("Connected to event server.");
-        source.onmessage = self.onMessage.bind(this);
-        source.onerror = (err) => {
-            console.log("EventSource failed:", err);
-            this.get('notificationStream').close();
-        };
-        
         self.set('source', source);
-        self.set('events', A([]));
     },
     
     willDestroyElement() {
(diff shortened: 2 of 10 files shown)


Xarthisius:
This looks great @bodom0015 !

BTW, I was able to get rid of duplicated notifications by applying the following:

```
diff --git a/app/services/notification-stream.js b/app/services/notification-stream.js
index cf03c83..8fd0721 100644
--- a/app/services/notification-stream.js
+++ b/app/services/notification-stream.js
@@ -83,8 +83,13 @@ export default Service.extend({
 
         // Push new event data
         let events = self.get('events');
-        if (event.json.type == 'wt_image_build_status') {
+        const found = events.some(
+            existingEvent => existingEvent.json._id === event.json._id && existingEvent.created === event.created
+        );
         // TODO: Handle updates to previous events
+        if (found) return;
+
+        if (event.json.type == 'wt_image_build_status') {
             events.unshiftObject(event);
             self.set('events', events);
             console.log("New event:", events);
```

bodom0015:
Good catch, @Xarthisius! The event stream was covering up the tooltips from the navbars, so I shifted thing around. I didn't realize that the z-index of that loading spinner was a problem again.

I went ahead and refactored that panel slightly to make the layout more sane (it's going away anyways in my other branch)

craig-willis:
This is great.  I have one comment and one question that will likely result in me creating a new task.

First, when I click on the (!) I get both the notifications and the tooltip.  I don't know whether there's anything that can be done about this, but since I'm not actually looking for the tooltip it seems odd. It could just be that the tooltip covers part of the notification window makes it more noticeable? 

Screen Shot 2019-05-09 at 10 25 27 AM

Second (and this is because I've been on Mars for a few weeks) -- when I run a Tale I seem to get the build complete notification before the building notification and, as expected, no notification when the Tale is actually running. Since the notification panel covers the launched tale, it's hard for me to know that it's running without clicking on it, but it's 90% obstructed by the notification panel. One solution would be to add a notification that the Tale is running or maybe there's a way to move the panel so that it isn't obstructing the "Launched Tales" panel?  
Screen Shot 2019-05-09 at 10 34 01 AM

bodom0015:
@craig-willis 

On your first point this behavior was intentional (see https://github.com/whole-tale/dashboard/pull/461/commits/d22525ff656901a7d322e5975e740f07e10d7f98), as the `notification-stream` component was covering all of the tooltips from the navbar links there. The tooltips should only appear when the mouse is hovering over the button, but we can have some discussion about what the proper behavior is if the button is only partially obscured, as it is in this case. If the button was fully obscured, I would agree that its tooltip should be obscured as well. Perhaps I should adjust the positioning of the stream?


On your second point, my understanding is that that has always been the case in the Run view...
 
> One solution would be to add a notification that the Tale is running

This sounds like the "restart notification" stuff we discussed earlier this week, if implemented correctly, could easily cover both of these cases.

>  or maybe there's a way to move the panel so that it isn't obstructing the "Launched Tales" panel?

That panel is going away in the very near future, so I haven't given it much thought... Any suggestion on where we might move the stream? Perhaps at the bottom-right, like the prior notifications from import tale? 

This may be another good point to bring up today at the UI/UX meeting, as I've already included a talking point about the stream in my notes.

craig-willis:
Both points taken and mine dropped.  With a real UI/X person involved, I can stop thinking about these things.

Issue #453, opened 2019-04-03
Add "Select Data" to "Create Tale" modal


craig-willis:
After completing #452, as per the mockups we will add the "Select Data" button to allow the user to select from registered datasets during Tale creation.

This includes:
* Adding the "Select Data" button to the modal
* Multiple selected datasets are displayed sensibly, possibly with a scrollbar.

As with #452, this needs to work with the "Analyze in WT" integration.

Xarthisius:
I'd suggest the following changes to the mockup:
1. Convert `Input Data` to a list.
2. Move `Input Data` label to a header / attached label.
3. Add plus (add data) functionally to the header / attached label.
4. Present selected datasets as a list
5. Move hint text to the tooltip

Sample mockup of that part:

![list](https://user-images.githubusercontent.com/352673/55503669-c40c8400-5614-11e9-96d7-112d9c3200d6.png)

Showing the tooltip:

![with_tooltip](https://user-images.githubusercontent.com/352673/55503672-c7a00b00-5614-11e9-9c00-32473439482a.png)

Xarthisius:
Alternatively to the tooltip, maybe a note at the top of the list?
![note](https://user-images.githubusercontent.com/352673/55504330-39c51f80-5616-11e9-8de2-556344af4c2e.png)

abrinckm:
maybe investigate #457 while I'm in this component working. But not to fix until after this issue has been worked on and PR resolved of course.

craig-willis:
Per [5/16 meeting](https://docs.google.com/document/d/1eC8_81IJlYGcvuuZLHDh6ePhide3BsQaudqk5hExIUo/edit), we will not implement select data on the create tale modal. We will support the parameters dataset ID (ala Analyze in WT) and link to documentation to describe how to add data after creating a tale.

Issue #452, opened 2019-04-03
Implement "Create Tale" modal


craig-willis:
Per the [mockups](http://www.crc.nd.edu/~kfurse/wholetale-css-mock-up-sui_f2018/browse.html), the "Compose" panel will be re-implemented as the "Create Tale" modal. This will require the following:

* Implementing the "Create Tale" modal
* Adding "Create New Tale" button. This can go on the Browse page for now until Browse is refactored.
* Removing the "Compose" panel

Note, the mockups don't factor in the "Analyze in WT" integration that prepopulates the form based on query string parameters, for example:

https://dashboard.stage.wholetale.org/compose?environment=RStudio+%28rocker%2Fgeospatial%29&name=Satellite+Telemetry+Dataset+%28Raw%29%3A+Juvenile+Bearded+and+Spotted+Seals%2C+2004-2006%2C+Kotzebue%2C+Alaska.&uri=10.24431%2Frw1k118

This will also need to be supported.

abrinckm:
![Select Data button](https://user-images.githubusercontent.com/8681111/55957870-5fb17c00-5c35-11e9-9fab-6eb7163e82cc.png)

Hi, 
Does anyone know what this button does? Does this open the select data modal?

abrinckm:
@craig-willis 
A couple of quick questions:

1. Do you want this button disabled and grayed out when a tale is being created?
2. What functionality do you expect when a user navigates to the compose link you copy pasted? Should that link read `/browse?[query params]` instead? And should this automatically trigger the "Create Tale" modal to open up with prepopulated values?

Thanks

abrinckm:
@craig-willis Another question:
The compose panel not only creates a tale but also automatically launches the tale. 
Do you want this modal to do the same?

Xarthisius:
1. No, it should be active. It's grayed out right now in the compose view, because there were some problems related to the code. As for the latter I think temporarily we should navigate to the browse view. The old Tale view is going away, the new Tale view is still being implemented (https://github.com/whole-tale/dashboard/pull/459).
[...]

Issue #366, opened 2019-01-28
Add Select Data modal to Compose page


craig-willis:
Per issue #362, we removed the Data panel from the Compose page. Next, we want to enable Select Data modal as per the [Create new tale mockups](http://www.crc.nd.edu/~kfurse/wholetale-css-mock-up-sui_f2018/browse-mytales.html). Initially, this could be either the usual "+" button or the "Select Data" button per the mockup.

Use case:
1. User composes a Tale, enters name and selects environment
2. User selects "Select Data" to pre-select the external data that will be included with the Tale
3. "Select Data" modal operates as with the external data tab.  Items are added to the list of input data.
4. User selects "Launch Tale" and the tale is created, instance launched, with selected data available via /data and in the external data tab as expected.

Note that we must maintain existing "Launch in WT" support.

bodom0015:
I think that we decided we weren't going to do this, right?

PR #465, opened 2019-04-23
Create tale modal

abrinckm:
This PR should cover the features described in issue #452

1. Implementing the "Create Tale" modal
1. Adding "Create New Tale" button. This can go on the Browse page for now until Browse is refactored.
1. Removing the "Compose" panel 
1. prepopulates the modal based on query string parameters 
    1. go to /browse?environment=RStudio+%28rocker%2Fgeospatial%29&name=Satellite+Telemetry+Dataset+%28Raw%29%3A+Juvenile+Bearded+and+Spotted+Seals%2C+2004-2006%2C+Kotzebue%2C+Alaska.&uri=10.24431%2Frw1k118
1. a dropdown button with a default action to create new tale and launch

It does not cover:
1. Does not track tale imports - see comments on issue #452
1. Does not implement the "Select data" button in the modal - it does nothing. That was a separate issue #453 and will be included in that PR and not this one.

To test: 
`git fetch origin`
`git checkout -b create_tale_modal origin/create_tale_modal`
Navigate to /browse page and click "Create New Tale".

For documentation on how to use this modal see:
[/app/components/create-tale-modal/README.md ](https://github.com/whole-tale/dashboard/blob/create_tale_modal/app/components/ui/create-tale-modal/README.md)

For an example on how to use this modal see:
[/app/templates/browse.hbs](https://github.com/whole-tale/dashboard/blob/create_tale_modal/app/templates/browse.hbs)


abrinckm:
@bodom0015 
I will commit some new code soon to disable the create button when the form is incomplete as you suggested.

 I cannot reproduce any of these problems:
* Only first creation will work without refreshing
* After running "Create Tale And Launch", I was brought to the Tale View page and clicking the "Back" button at the top-left of the Tale View UI brought me to a half-rendered Browse view, which threw an error in the console and the rest of the view did not render:
* When the developer console is open, the modal's left: 50% styling pushes it off the screen to the right

Are you running against a local development stack or the girder.dev?
What browser are you using? 

For the following issues I will leave a comment to open some discussion about what may be done:
* max-height seems weird... 
This is a design choice made by the UX team and Kristina. The height and max height are directly copied from the styles provided from Kristina. What do you suggest I should do about this? How much do you want the modal to expand?

* Import Tale fails silently if you are at the limit for instances (may be out of scope for this PR):
  I've stated both in this PR and in the code and also as a comment in issue #452 that tracking jobs are not handled in this PR. This includes failed jobs. I would request based on the discussion with Kacper that we first gather requirements from the UX/UI team for how to handle this and then make the task of "tracking import jobs" a separate task from how this modal works.

* Modal closes on errors 
Yes. I'm using the same error modal your team implemented for the compose page. May I suggest using a toast message instead of a modal? This might work. Because I cannot have two modal open simultaneously, I must close the first modal to open the error modal.

* After importing a Tale, navigating away from Browse and back again preserves the query string and re-opens the Modal
I know that a solution is being discussed in another issue about this. What was the final decision? what do you think should be done?

bodom0015:
> Are you running against a local development stack or the girder.dev?
I'm using a full [deploy-dev stack](https://github.com/whole-tale/deploy-dev), all running the latest `master` except for dashboard running this branch

> What browser are you using?
Google Chrome Version 73.0.3683.103 (Official Build) (64-bit)

> For the following issues I will leave a comment to open some discussion about what may be done:
> 
> * max-height seems weird...
> * Import Tale fails silently if you are at the limit for instances (may be out of scope for this PR):
It's fine to leave these both for now - we can address later after getting the take from the UX team

> * Modal closes on errors
>   Yes. I'm using the same error modal your team implemented for the compose page. May I suggest using a toast message instead of a modal? This might work. Because I cannot have two modal open simultaneously, I must close the first modal to open the error modal.
Semantic UI actually does allow for [multiple modals](https://semantic-ui.com/modules/modal.html#multiple-modals), but I agree with your point that this is less than desirable... I don't think the "error modal" pattern was intended to be used in conjunction with other modals, but you are right in pointing out that we don't have a consistent pattern for such things.

> * After importing a Tale, navigating away from Browse and back again preserves the query string and re-opens the Modal
>   I know that a solution is being discussed in another issue about this. What was the final decision? what do you think should be done?
[Ember's Query Parameter documentation](https://guides.emberjs.com/v3.1.0/routing/query-params/#toc_transitionto) describes a pattern for changing the `queryParams` without actually navigating.

As long as the router is defined in the component:
```javascript
   router: service(),
```
we should be able to return them to default values using something like the following:
```javascript
       this.router.transitionTo({ queryParams: { environment: null, name: null, uri: null, ... }});
```

This could be done either when the modal opens or (preferably) when the modal closes - this way, if the user refreshes the page with the modal open, the modal re-opens, but otherwise remains closed after a refresh.

abrinckm:
@bodom0015  Thank you for your quick response!

I'm still not able to reproduce the errors. Let's try and drill down to what the problem might be. 

I'm on the dev branch of the deploy-dev (I had problems on master, Kacper asked me to use dev)
I'm using node v8.11.4 
ember-cli v3.14
semantic-ui v3.0.3

Also, can you give me the exact values that are being 
> Semantic UI actually does allow for multiple modals

Okay. Let me see if I can try to do this.

> Ember's Query Parameter documentation describes a pattern for changing the queryParams

I'm going to push a commit to do this soon.

bodom0015:
We shouldn't have any version conflicts if we're both using [deploy-dev](https://github.com/whole-tale/deploy-dev/tree/dev). The dashboard's `package-lock.json` file (generated by NPM) should lock our downloaded dependencies to known versions, and deploy-dev uses Docker to provide a known version of NodeJS...

In any case, here's what I have:
* NodeJS v8.15.1
* ember-cli@3.14
* semantic-ui-css@2.3.1
* semantic-ui-ember@3.0.3

________________________________________________________________

> After running "Create Tale And Launch", I was brought to the Tale View page and clicking the "Back" button at the top-left of the Tale View UI brought me to a half-rendered Browse view, which threw an error in the console and the rest of the view did not render

Demonstration: http://recordit.co/ZW8uxoMB5L

Steps to Reproduce:
1. Open "Create Tale" modal
2. Enter a name and choose an environment
3. Click "Create Tale And Launch"
    * You should be brought to the Tale View page
    * NOTE: in my case, I am receiving an Exception here from within a Promise
4. Click the "Back" button at the top-left of the Ember view (e.g. **NOT** your browser's back button)
    * You should be brought back to the Browse view

Actual Results:
* The Browse view is only half-rendered, and shows no tales listed, no `tale-instances` panel on the right, etc

Expected Results:
* The Browse view should render and be fully-functional

________________________________________________________________

> Only first creation will work without refreshing

I am having issues reproducing this one reliably as well, but I still see it pop up occasionally. See GIF below.

Demonstration: http://recordit.co/QNVJhoviIF

Steps to Reproduce:
??? (still working on how to reproduce reliably)

________________________________________________________________

> When the developer console is open, the modal's left: 50% styling pushes it off the screen to the right

Demonstration: http://recordit.co/PCykpJjkQC

Explantation: The `left: 50%` styling may only be an issue for smaller screens/viewports

Steps to Reproduce: 
1. Open developer console
2. Open the "Create Tale" modal
3. Grab the top of the dev console and drag it up and down to resize the viewport
    * With a small enough viewport, it appears that `left: 50%` style is applied

abrinckm:
I can't reproduce the error you are seeing when creating a tale. 

http://recordit.co/fzGT4qyCji

However, the video was extremely helpful. It showed that at least for you, the problem originates from a rejected promise. Which is actually a failed api call. I will try to seed such an error to mimic the problem and then see if I can reproduce what you experience. 

You said you were using the master branch of the deploy-dev stack. I was not able to get that to work so I am using the dev branch suggested by Kacper.

Can you provide for me one last piece of information? Can you go into your network tab and see what api call has failed for you and let me know? It will help me mimic the problem. 

------

I cannot reproduce the modal positioning issue even on lower resolution.

http://recordit.co/eSNtxztW4h

My modal doesn't use a `left: 50%` styling and I haven't applied any positioning to the modal, meaning it's using the same positioning rules set for all other modals in the dashboard. 

------

I was able to successfully reproduce the dataSet is null issue. 
May I request that we punt on this particular issue since the select data button has not been implemented yet? This button will be implemented as part of a separate PR. Once done, this bug will not be a problem.

bodom0015:
> I can't reproduce the error you are seeing when creating a tale.
> 
> http://recordit.co/fzGT4qyCji
> 
> However, the video was extremely helpful. It showed that at least for you, the problem originates from a rejected promise. Which is actually a failed api call. I will try to seed such an error to mimic the problem and then see if I can reproduce what you experience.
> 
> You said you were using the master branch of the deploy-dev stack. I was not able to get the master branch to work so I am using the dev branch of the deploy-dev stack instead as suggested by Kacper.
> 
> Can you provide for me one last piece of information? Can you go into your network tab and see what api call has failed for you and let me know? It will help me mimic the problem.

I'm not seeing any failures from the networks tab, but was given a different error message this time:
```
"Error: Attempted to handle event `didSetProperty` on <tale:null> while in state root.loading. Called with {name: icon, oldValue: undefined, originalValue: undefined, value: /images/whole_tale_logo-dbfba3ae0a521b126685c8104cf64020.png}.
    at new n (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:1867:13)
    at e._unhandledEvent (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:10923:121)
    at e.send (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:10912:19)
    at o.set (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:11330:120)
    at Re.je._set (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:2419:138)
    at Re.je.setWithSuspend (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:2419:17)
    at Re.je.set (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:2417:187)
    at Oe (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:2392:34)
    at o.set (https://dashboard.local.wholetale.org/assets/vendor-7883873bae446edf359150c5e1f6357b.js:3025:89)
    at https://dashboard.local.wholetale.org/assets/wholetale-eb9f6e9461873147fa1ef86dd89bddce.js:397:263"
```

This appears to be similar to the error seen in issue #395.. I don't think this issue was ever resolved

______________________________________________________________

> I cannot reproduce the modal positioning issue even on lower resolution.
> 
> http://recordit.co/eSNtxztW4h
> 
> The modal doesn't use a `left: 50%` styling and I haven't applied any positioning to the modal, meaning it's using the same positioning rules set for all other modals in the dashboard.

That makes sense... since we've seen it with other modals it might just be the same issue we've seen there before. No need to worry about this one for now.

______________________________________________________________

> I was able to successfully reproduce the dataSet is null issue.
> May I request that we punt on this particular issue since the select data button has not been implemented yet? This button will be implemented as part of a separate PR. Once done, this bug will not be a problem.

I would be more inclined to agree if it wasn't a 1-line change... This seems like a mistake with a simple fix. I'm not worried about `dataSet` being functional, but it also shouldn't force the user (us) to reload the page if the wrong button is clicked.

Does changing line 56 [here](https://github.com/whole-tale/dashboard/pull/465/files#diff-e516ddfabad21b56443b541e2052d25dR56) from `null` to `[]` introduce a merge conflict with another branch?

app/components/ui/create-tale-modal/component.js
```diff
@@ -0,0 +1,157 @@
...
+    },
+
+    // ---------------------------------------------------------------------------------
+    // clears all text fields 
+    // ---------------------------------------------------------------------------------
+    clearModal() {
+      this.set('importing', false);
+      this.set('datasetAPI', null);
+      this.set('imageId', null);
+      this.set('dataSet', null);
```
bodom0015:
```suggestion
      this.set('dataSet', []);
```

abrinckm:
Hi @bodom0015. Thank you! 

I've read your comments. I will resolve these issues likely tomorrow if not Monday. I'm out of town today.

abrinckm:
@bodom0015 
I still can't reproduce that error. I've tried mimicking this by seeding my own errors but still can't get this to fail. Perhaps I'm seeding in the wrong place?

A few more questions, but we may need to have a screen share session and work through this on your machine to solve it since I can't reproduce this on my end.

What did you do to solve the first error that isn't happening anymore? What changed to cause this new error?
Does this error only occur when navigating to an existing tale view and then navigating back to browse?
Or does this error only occur when creating a new tale?

bodom0015:
> I still can't reproduce the errors you're seeing in the console. I've tried mimicking this by seeding my own errors but still can't get this to fail. Perhaps I'm seeding in the wrong place?

I'm not entirely sure what you mean by "seeding errors"... You mean placing breakpoints? Or maybe are you attempting to simulate a Tale Launch failure by killing repo2docker or something similar? 

> A few more questions, but we may need to have a screen share session and work through this on your machine to solve it since I can't reproduce this on my end.

We can screen share on this, if needed. I'd like to also have @craig-willis or @Xarthisius take a look at this, if possible.. if it's only broken on my machine then that's something I can dig into.

> What did you do to solve the first error that isn't happening anymore? What changed to cause this new error?

All I've done is pulled from your branch to get your new commit(s) to disable the submit button.. no manual code changes have been made.

> Does this error also occur when navigating to an existing tale view and then navigating back to browse? Or does this error only occur when creating a new tale?

* Working: I am able to "View" and return to the Browse view just fine
* Working: I am able to "Create Tale" and return to the Browse view
* Not working for me: Everytime I run "Create Tale And Launch", I am unable to return to Browse without refreshing - this is true regardless of whether or not I already have 2 running Tales

app/components/ui/create-tale-modal/component.js
```diff
@@ -0,0 +1,158 @@
...
+    const store = this.get('store');
+    let newTale = store.createRecord('tale', {title, imageId, dataSet, config});
+
+    try {
+      let item = await newTale.save();
+      let taleId = item.id;
+      
+      if (this.createAndLaunch) {
+        let newInstance = this.get('store').createRecord('instance');
+        newInstance.save({adapterOptions:{queryParams:{imageId, taleId}}});
```
bodom0015:
```suggestion
        await newInstance.save({adapterOptions:{queryParams:{imageId, taleId}}});
```

app/components/ui/create-tale-modal/component.js
```diff
@@ -0,0 +1,158 @@
...
+  //   createNewTaleButton() = Event: onClick "Create New Tale" button
+  //   clearModal()          = Event: onHide modal 
+  //   setModalFromQueryParams() = Event: onShow modal
+  //   setDefaultAction()    = Event: onSelect "Create New Tale" dropdown
+  // ---------------------------------------------------------------------------------
+  actions: {
+    // ---------------------------------------------------------------------------------
+    // Either imports tale from a dataset or creates a new tale
+    // ---------------------------------------------------------------------------------
+    async createNewTaleButton() {
```
bodom0015: I don't think `async` is needed here, since we aren't calling `await` and neither of the functions called below will return a Promise.

app/components/ui/create-tale-modal/component.js
```diff
@@ -0,0 +1,158 @@
+import Ember from 'ember';
+import Component from '@ember/component';
+import {inject as service} from '@ember/service';
+import {later} from '@ember/runloop';
```
bodom0015: Lint errors:
```javascript
/usr/src/node-app/app/components/ui/create-tale-modal/component.js
   12:3   error  Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties  ember/avoid-leaking-state-in-ember-objects
   13:3   error  Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties  ember/avoid-leaking-state-in-ember-objects
   16:21  error  Use import { computed } from '@ember/object'; instead of using Ember.computed                          ember/new-module-imports
   20:13  error  Use import { computed } from '@ember/object'; instead of using Ember.computed                          ember/new-module-imports
   58:7   error  Do not use global `$` or `jQuery`                                                                      ember/no-global-jquery
   58:7   error  '$' is not defined                                                                                     no-undef
   79:11  error  Do not use global `$` or `jQuery`                                                                      ember/no-global-jquery
   79:11  error  '$' is not defined                                                                                     no-undef
   84:19  error  Do not use global `$` or `jQuery`                                                                      ember/no-global-jquery
   84:19  error  '$' is not defined                                                                                     no-undef
  104:15  error  Do not use global `$` or `jQuery`                                                                      ember/no-global-jquery
  104:15  error  '$' is not defined                                                                                     no-undef
  150:11  error  'job' is assigned a value but never used                                                               no-unused-vars

✖ 13 problems (13 errors, 0 warnings)
```

Xarthisius:
FWIW, I was able to reproduce the issue that Mike mentioned following the steps in https://github.com/whole-tale/dashboard/pull/465#issuecomment-487692927:
bug01

app/components/ui/create-tale-modal/component.js
abrinckm: Thanks. The dashboard has many linter errors that can be ignored because we never configured any convention.

app/components/ui/create-tale-modal/component.js
abrinckm: Correct. It used to have code in there but was split up into two different functions.

app/components/ui/create-tale-modal/component.js
abrinckm: No. It does not. await is used if you want to capture the result of the promise. We are not capturing anything here.

abrinckm:
@Xarthisius @bodom0015 
Sorry. You are overwhelming me with messages in this pull request. I am unable to keep up with the work here...

abrinckm:
@bodom0015 
Please create another list of checkboxes of things that need to be fixed in this pull request. The list keeps getting larger and larger and I'm losing track of it all...

abrinckm:
@bodom0015 You asked a question before somewhere in this thread about "seeding errors" http://www.professionalqa.com/error-seeding

@Xarthisius Thank you for verifying you can reproduce. I cannot. I welcomed a screen share session.

abrinckm:
@bodom0015 You asked in this thread about async/await. It is preferred over promise chains. You can read here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

bodom0015:
@abrinckm Thank you for fixing the high-priority issues, and I apologize for being so overwhelming. 

Please fix this last issue, and then I think we can approve this one and move on:
- [ ] Resolve merge conflicts

______________________________

**I'm also noting the following here for completeness' sake.**

There are already issues filed for the following:
- Modal positioning: #174 / #271 / etc
- "Import Tale" progress: #366 - or whatever future issue includes progress updates for "Import Tale"

I can file new issues for the following:
- "Create Tale And Launch" error - `tale.view` is going away in the near future anyways
- Modal closes on error - minor problem, may be a more general pattern problem

We can investigate patterns surrounding the async/await pattern as time goes on

Xarthisius:
@abrinckm could you resolve the conflicts on this PR? I'm gonna do a 2nd pass on review, and per @bodom0015 's comment merge and file any additional things I find as separate issues. TIA!

craig-willis:
@abrinckm I'm testing this PR for the first time and seeing several unexpected behaviors:

* As reported by both @bodom0015  and @Xarthisius , "Create and Launch" takes me to the "View" page and clicking "Back" takes me to a broken Browse page (same as above).  I expect to remain on the browse page as the instance starts.
* Environment dropdown shows only 2 entries, requiring me to scroll to scroll the modal to find the others. As a user, I don't expect to need to scroll to find dropdown items.
* If I have two tales running and try to "Create and Launch" a third, I don't see any error.
* When following the "Analze in WT" path (browse?environment=...&name=...) no new tale appears on the browse page and the instance does not appear to start. Both appear only after refresh.

If you need me to demonstrate any of these, just find me on Slack.

abrinckm:
@Xarthisius @craig-willis @bodom0015 
I'd like to propose that we cancel and delete this branch. 
It will be less time to start fresh and code a new modal after all of the refactoring is complete.
As opposed to trying to get this working with current browse page and tale view now, and then working again with refactored browse page later. 
Also, Mike was going to create separate issues that were posted in this PR, which also takes away his time.

What do you guys think?

bodom0015:
Sorry that I missed this, but if that might make things easier that would work for me. 

Surely we could at least reuse the some of the `create-tale-modal` component/template and styling introduced in this branch, so that you don't have to start entirely from scratch?

abrinckm:
- [x] remain on the browse page as the instance starts.
- [ ] As a user, I don't expect to need to scroll to find dropdown items.
- [ ] "Create and Launch" a third should display error
- [ ] When following the "Analze in WT" path (browse?environment=...&name=...), somehow get tale and display launching

abrinckm:
@craig-willis see the comment I made 14 days ago on this thread with the checkboxes. you can see progress.

last checkbox cannot be checked (see also the very first comment I made on this PR on April 23rd).  I'm not familiar with the "job" stuff, but I'm guessing the browse panel needs refactored to subscribe to "job" notifications and listen for instance creation. I don't know if this is even supported from the backend. we can talk on the meeting.

ThomasThelen:
I think that this is ready for review again. There may be a few minor issues lingering, but I think these are small enough to be tackled as individual issues.
The one that I've seen is that the `Create New...` menu opens when the modal closes on a create new tale action.

bodom0015:
These can be handled as separate issues if needed, but I am enumerating issues as I find them:
1. The issue pointed out by @ThomasThelen above - the dropdown reopens after the modal closes
2. Tales do not appear under "Browse" after creation without a refresh
3. Delete tales appears to be broken: `TypeError: Cannot read property 'name' of undefined` coming from the `delete-modal`


Solutions:
1. I believe that the issue that @ThomasThelen mentions above can be solved by adding a `showOnFocus:false` parameter. Changing the init logic to the following (in `browse/left-panel/component.js:L23`) appeared to prevent the dropdown from opening when it regains focus after the modal disappears:
```
    $('.ui.dropdown').dropdown({
      action: 'select',
      showOnFocus:false
    });
```
2. Note that by implementing #508 we can likely hide this bug without actually needing to fix it, as staying on the Browse after creating a Tale will no longer be a supported case
3. This was likely caused by a missed merge fix as well as the same bug discovered by @Xarthisius in #514 - pulling from master and repairing conflicts should fix this issue. From `app/components/ui/tale-browser/component.js:L260`, this line also needs to change from:
```
        component.updateModels(component, component.get('models'));
``` 
to instead be:
```
        component.updateModels(component, component.get('models').tales);
```

craig-willis:
I see two issues, but recommend that we get this merged and treat them as separate issues.

1. As noted in https://github.com/whole-tale/dashboard/pull/465#issuecomment-514377654 item 2, the tale doesn't appear without refresh.  PR #508 fixes this only in the Create and Launch case, but not in the Create without launch case. 
2. For the [Analyze in WT](https://dashboard.local.wholetale.org/browse?environment=RStudio+%28rocker%2Fgeospatial%29&name=Satellite+Telemetry+Dataset+%28Raw%29%3A+Juvenile+Bearded+and+Spotted+Seals%2C+2004-2006%2C+Kotzebue%2C+Alaska.&uri=10.24431%2Frw1k118) case, the dataset information isn't displayed in the dialog, only the placeholder text to add data after launching.  I expect the same behavior as on production (e.g., "Data Source: 10.24431/rw1k118") or similar.

Issue #466, opened 2019-04-24
Re-open Publish modal on auth=true


ThomasThelen:
The publishing modal used to open when the `auth` queryParam was true-it looks like this isn't happening any more. This issue is to make the modal pop back up.

ThomasThelen:
This has been merged. Closing.

PR #467, opened 2019-04-25
Open the publish modal when auth=true

ThomasThelen:
Fixes issue #466 

This small change opens the publishing modal when the `auth` queryParam is true

To test:
1. Check `fix_auth_popup` out
2. Navigate to the run page
3. Sign out of DataONE
4. Open the publish modal
5. Sign in
6. After you are re-directed back to whole tale, note `auth=true` in the URL
7. The dialog should open

### DIFF
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -116,7 +116,8 @@ export default Component.extend(FullScreenMixin, {
           let queryParams = this.get('params')
           if (queryParams) {
             if (queryParams.auth === 'true') {
-              this.send('openPublishModal', this.model.taleId)
+              this.router.transitionTo({ queryParams: { auth: null }});
+              this.send('openPublishModal', this.model.taleId);
             }
           }
         });
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -14,7 +14,7 @@ const O = EmberObject.create.bind(EmberObject);
 export default Component.extend(FullScreenMixin, {
     layout,
     classNames: ['run-left-panel'],
-    router: service('-routing'),
+    router: service(),
     internalState: service(),
     apiCall: service('api-call'),
     dataoneAuth: service('dataone-auth'),
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -5,7 +5,7 @@ import config from '../../../../config/environment';
 import { scheduleOnce, later, cancel } from '@ember/runloop';
 import EmberObject, { computed } from '@ember/object';
 import { A } from '@ember/array';
-import { not } from '@ember/object/computed';
+import { not, alias } from '@ember/object/computed';
 import $ from 'jquery';
 import layout from './template';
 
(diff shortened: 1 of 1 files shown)


ThomasThelen:
Nice catch. I also see that the canceling behavior is off after navigating back to /run from another page (like browse). I'll not that I only see this behavior when auth=true, I click `Browse` or any other tab, and then navigate back to run.

Interestingly enough, other parts of the modal act different too. For example, `onRepositoryChange` will set `respText` to the array of publishing locations. If you change the repository after a page refresh things work as expected.

From [the docs](https://guides.emberjs.com/release/routing/query-params/#toc_sticky-query-param-values),
If you wish to reset a query param, you have two options:

1.explicitly pass in the default value for that query param into link-to or transitionTo.
2.use the Route.resetController hook to set query param values back to their defaults before exiting the route or changing the route's model.

1. This might be able to be done by catching page changes to /run, calling `transitionTo` the run route (which would be weird since we're already heading there), and then resetting the query param.
 
2. It looks like modifying queryParams is done via the controller, which is unfortunate since we typically go with component/templates instead. In the run page component , I use the `-router` service so that I can see the queryParams.

3. One hackish way that would probably avoid the controller is checking the last page in the history. If the last page wasn't ORCID, don't open the modal.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -110,12 +112,13 @@ export default Component.extend(FullScreenMixin, {
...
-            // If ?auth=true
-            // Open Modal
-            const modalDialogName = 'ui/files/republish-modal';
-            this.showModal(modalDialogName, this.get('modalContext'));
-        
+          // Check if we're coming from an ORCID redirect
+          let queryParams = this.get('params')
+          if (queryParams) {
+            if (queryParams.auth === 'true') {
+              this.send('openPublishModal', this.model.taleId)
```
bodom0015:
```suggestion
              this.router.transitionTo({ queryParams: { auth: 'false' }});
              this.send('openPublishModal', this.model.taleId);
```

bodom0015: For the above to work, I think line 17 (above) in this file also needs to change from
```javascript
    router: service('-routing')
```
to
```javascript
    router: service()
```

bodom0015: One other thing to note, setting `auth` to `null` here will actually clear the query string parameter from the URL - this might be a better pattern than setting it to explicitly to `false`.

ThomasThelen: Hmm I haven't been able to get this to work. The closest thing I saw was doing from [inside the route](https://github.com/emberjs/ember.js/blob/v3.9.1/packages/%40ember/-internals/routing/lib/system/route.ts#L661). 

I put this logic in the route, and also couldn't get it to reset the queryParams? I tried it with transitionTo and by manually setting them via this.set and didn't have any luck there.

bodom0015: Hmmm... that's definitely strange.

All I needed was the following, and it appeared to work (in Chrome, at least):
```javascript
    router: service(),      // note: this is different from service('-routing')

    didInsertElement() {
        scheduleOnce('afterRender', this, () => {

            // Check if we're coming from an ORCID redirect
            let queryParams = this.get('params')
            if (queryParams && queryParams.auth === 'true') {
                this.router.transitionTo({ queryParams: { auth: null }});
                this.send('openPublishModal', this.model.taleId);
            }
        });
    },
```

You said that this doesn't clear the `auth` query string param for you?

Xarthisius:
@ThomasThelen I see that you merged this PR, was that intentional? I see that @bodom0015 requested some changes, but didn't have a chance to do a 2nd pass. I haven't review it yet at all.

ThomasThelen:
Oops @Xarthisius I wasn't sure if you wanted to take a look.  Here's a gif of the final product after the requested changes were made.

HhwjJBS4o2

Xarthisius:
Yeah, sorry I was slow to review lately, but I'm trying to pick up the slack.

PR #474, opened 2019-05-03
Add production flags to dataone repositories

ThomasThelen:
We need to tell whether we're publishing to production or dev DataONE. This PR 

1. Sends the user to prod/dev DataONE login
2. Uses the right prod/dev JWT
3. Lets the backend know whether it's prod/dev

The biggest UI change is that the user is prompted to login to DataONE _after_ they select a repository. The reason for this is that dev and production dataone uses two different login servers (and we need to know which one to send the user to).
To Test:
See [this issue](https://github.com/whole-tale/gwvolman/pull/72)

Note: I accidentally published a few Tales to production using this, so be careful when playing with the repository dropdown.

### DIFF
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -162,8 +162,6 @@ export default Component.extend(FullScreenMixin, {
      * @method createTooltips
      */
     createTooltips() {
-        console.log('Creating tooltips...');
-        
         // Create the popup in the main title
         $('.info.circle.blue.icon.main').popup({
           position: 'right center',
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -214,16 +214,6 @@ export default Component.extend(FullScreenMixin, {
         });
     },
     
-    getRepositoryFromName(name) {
-        // Given a repository name, find the membernode URL
-        let repositoryList = this.get('repositories');
-        for (var i = 0; i < repositoryList.length; i++) {
-          if (repositoryList[i].name === name) {
-            return repositoryList[i];
-          }
-        }
-    },
-    
     handlePublishingStatus(tale, jobId) {
         let self = this;
         let currentLoop = null;
@@ -337,10 +327,16 @@ export default Component.extend(FullScreenMixin, {
           let url = `${config.apiUrl}/tale/${id}/export`;
           window.location.assign(url + '?token=' + token + '&taleFormat=' + format);
         },
-        
-        authenticateD1(taleId, coordinatingNode) {
-          let callback = `${this.get('wholeTaleHost')}/run/${taleId}?auth=true`;
-          let endpoint = this.dataoneAuth.getPortalEndpoint(coordinatingNode)
+
+        /**
+        * Redirect to the selected repository's authentication portal.
+        *
+        * @method authenticateD1
+        * @param instanceId The ID of the running instance
+        */
+        authenticateD1(instanceId) {
+          let callback = `${this.get('wholeTaleHost')}/run/${instanceId}?auth=true`;
+          let endpoint = this.dataoneAuth.getPortalEndpoint(this.selectedRepository.coordinatingNode)
           endpoint += '/oauth?action=start&target=';
           window.location.replace(endpoint + callback);
       },
@@ -348,23 +344,19 @@ export default Component.extend(FullScreenMixin, {
         openPublishModal(tale) {          
             this.resetPublishState();
             this.set('taleToPublish', tale);
-            this.set('selectedRepositoryName', null);
             const self = this;
             $('#publish-modal').modal({ 
                 onApprove: () => false,
                 onDeny: () => false,
-                onVisible: () => { self.set('selectedRepositoryName', self.get('repositories')[0].name); }
+                onVisible: () => { self.set('selectedRepository', self.get('repositories')[0]); }
             }).modal('show');
         },
         
         submitPublish(tale) {
             const self = this;
-            console.log('Now publishing:', tale);
 
-            let targetRepo = self.get('selectedRepositoryName');
-            let repository = self.getRepositoryFromName(targetRepo);
-
-            let dataOneJWT = this.dataoneAuth.getDataONEJWT(repository.coordinatingNode)
+            let repository = this.selectedRepository;
+            let dataOneJWT = this.dataoneAuth.getDataONEJWT(repository.coordinatingNode);
             if (!dataOneJWT) {
               // reroute to auth
               $('#dataone-auth-modal').modal('show');
@@ -399,13 +391,17 @@ export default Component.extend(FullScreenMixin, {
             $('#publish-modal').modal('hide');
             this.resetPublishState();
         },
-        
+
+        /**
+        * Called when the user selects a repository in the dropdown menu.
+        *
+        * @method onRepositoryChange
+        */
         onRepositoryChange: function () {
-          // Called when the user changes the repository
-          let respText = $('.repository.selection.dropdown.ui.dropdown').dropdown('get text')
-          this.set('selectedRepositoryName', respText);
-          let repository = this.getRepositoryFromName(respText);
-          this.set('selectedRepository', repository)
+          let repositoryName = $('.repository.selection.dropdown.ui.dropdown').dropdown('get text');
+          console.log('Selected '+ repositoryName + ' for publishing');
+          let repository = this.get('repositories').find((repo) => repo.name === repositoryName);
+          this.set('selectedRepository', repository);
         },
     }
 });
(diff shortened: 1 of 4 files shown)


bodom0015:
> Note: I accidentally published a few Tales to production using this, so be careful when playing with the repository dropdown.

Do DataONE's production repositories allow you to delete or archive published artifacts (or somehow hide them from search)? I'm happy to test out the dropdown presentation and everything, but am struggling to find a way to test the functionality here without polluting their production resources with dummy files.

Any advice on how best to test this?

craig-willis:
@ThomasThelen  As discussed on the dev call today, the recommendation for v0.7 is to replace the `isProduction` flag with the CN URL for each repository and pass it through instead.  This should mean that you can drop the hardcoded CNs and CN check logic in `getDataONEJWT` and `getEndpoint`.  

@bodom0015  Tommy and I agreed that we would publish a legitimate Tale as our test.  This will likely be some variation of the "Water Tale" that we could reasonably see ourselves referencing via DOI.  It is possible for a DataONE administrator to delete the package after-the-fact, but not recommended.

Other notes:
* Q. Should production users be able to publish to dev?  This is apparently recommended by DataONE for users to test the publication project without polluting the production repository (similar to Zenodo sandbox).  This seems like a reasonable capability to support.  However, we do not want dev instances to support publishing to production.

ThomasThelen:
I've made the appropriate changes to send the coordinating node to the backend in ca4c565459d38c532ac19e2557d4569ce1c31a20.
In the process you'll see that I renamed a few variables to make it more clear which node we're dealing with. I also renamed `getEndpoint` to `getPortalEndpoint`.

app/components/dashboard/run/left-panel/component.js
```diff
@@ -361,38 +340,44 @@ export default Component.extend(FullScreenMixin, {
...
+                onVisible: () => { self.set('selectedRepositoryName', self.get('repositories')[0].name); }
             }).modal('show');
         },
         
-
         submitPublish(tale) {
             const self = this;
             console.log('Now publishing:', tale);
+
+            let targetRepo = self.get('selectedRepositoryName');
```
craig-willis: Very minor observation, but couldn't you just use the selected index instead of getting the name and using a separate function to lookup by name?

craig-willis:
@ThomasThelen if we don't want developers to publish to DataONE production, then we need some sort of configuration to determine whether to display the prod option, right?

app/components/dashboard/run/left-panel/component.js
```diff
@@ -361,38 +340,44 @@ export default Component.extend(FullScreenMixin, {
...
-        openPublishModal(tale) {
-            const jwt = this.getDataONEJWT();
-            if (!jwt) {
-                // reroute to auth
-                $('#dataone-auth-modal').modal('show');
-                return;
-            }
-            
-            this.set('dataoneJWT', jwt);
+        authenticateD1(taleId, coordinatingNode) {
```
craig-willis: I may be misunderstanding something, but `authenticateD1` seems to be called before I select the member node to publish to.  This means that the coordinatingNode is always the staging node, since the dropdown defaults to the first entry (dev). As discussed on the dev call, this would work since the JWT is raw ORCID and identical for stage/prod. But, for this to work as intended it seems that the user would need to select the member node before authenticating, right?

ThomasThelen:
> @ThomasThelen if we don't want developers to publish to DataONE production, then we need some sort of configuration to determine whether to display the prod option, right?

We can hide it if we want by using the dev flag in the environment, but I think everyone knows to not publish to production. It might actually be nice to have a confirmation popup, warning that you won't be able to delete the published Tale.

app/components/dashboard/run/left-panel/component.js
ThomasThelen: The semantic ui dropdown doesn't have a method for retrieving the index, but it does for text https://semantic-ui.com/modules/dropdown.html#/usage Scroll down to `Behavior`.
I use `get text` [here](https://github.com/whole-tale/dashboard/blob/ca4c565459d38c532ac19e2557d4569ce1c31a20/app/components/dashboard/run/left-panel/component.js#L405), every time the repository is changed. I made [getRepositoryFromName](https://github.com/whole-tale/dashboard/blob/ca4c565459d38c532ac19e2557d4569ce1c31a20/app/components/dashboard/run/left-panel/component.js#L219) as a way to get from the name, to the repository.


In 366 I could have used self.get('selectedRepository') and saved the call on 367 since it should be set by the time this code is reached.

app/components/dashboard/run/left-panel/component.js
ThomasThelen: `authenticateD1` gets called when the user [clicks Continue to DataOne Login](https://github.com/whole-tale/dashboard/blob/ca4c565459d38c532ac19e2557d4569ce1c31a20/app/components/dashboard/run/left-panel/template.hbs#L102) after checking the token endpoint. I took out the call to the publish endpoint and confirmed that the correct nodes are checked on development and on production.

In the link above, `selectedRepository` is set each time the repository is changed, and should be current when this code is hit.

craig-willis: Clearly user error on my part -- maybe a stale dashboard build.  I see now that the login prompt appears after I hit "Publish" (after selecting the target node) and that the correct CN is checked for the token.  During my testing, the login prompt was appearing when I clicked the "Publish Tale..." menu, but that's changed.

craig-willis:
I'd like for @mbjones  to weigh on whether to disable publish to production in our dev and staging instances.  I disagree that everyone knows not to.  Developers, PIs and other stakeholders tinkering with dev or staging instances will not necessarily know that they shouldn't.

app/components/dashboard/run/left-panel/component.js
bodom0015: You can actually use the built-in prototype method `Array.find()` for this too.

`let selectedRepo = repositories.find((repo) => repo.name === selectedRepositoryName);` would do the same thing without needing to write/maintain a custom helper function for it.

app/components/dashboard/run/left-panel/template.hbs
```diff
@@ -126,7 +126,7 @@
       <div class="ui form" style="margin-bottom:15px;">
           <div class="required inline field">
               <label>Please choose a target repository:</label>
-              {{#ui-dropdown class=repoDropdownClass allowAdditions=false selected=selectedRepository onChange=(action 'onRepositoryChange')
+              {{#ui-dropdown class=repoDropdownClass allowAdditions=false selected=selectedRepositoryName onChange=(action 'onRepositoryChange')
```
bodom0015: Forgive me, but I'm not understanding the changes from ` selectedRepository` to `selectedRepositoryName` here.

If we used to select the entire repository object with the dropdown, then why change it to select only the name if we would then need to immediately do a lookup by name to get the full object? Am I just overlooking something obvious?

mbjones:
Yes, please do not allow publication to production from dev environments.  People will mistakenly publish things there.  We've already had to manage a few deletions from Tommy.  Can't the list of endpoints be a configuration setting and set up differently for each environment?

ThomasThelen:
In regards to 035fd7f, `ENV.dev` was getting set to 'true' or 'false', but as strings instead of boolean type. That commit will turn the value into a bool, which was causing trouble in 
8012834

app/components/dashboard/run/left-panel/component.js
```diff
@@ -110,38 +109,30 @@ export default Component.extend(FullScreenMixin, {
...
           if (queryParams) {
             if (queryParams.auth === 'true') {
-              this.router.transitionTo({ queryParams: { auth: null }});
-              this.send('openPublishModal', this.model.taleId);
-            }
+              this.get('store').findRecord('tale', this.model.taleId, {
+                reload: true
+              })
+              .then(resp => {
+              this.send('openPublishModal', resp)
```
bodom0015: Tale should already be available via the `this.model.tale` (and will be `this.model` after the 0.8 refactoring). Was there a reason to requery for it here?

app/components/dashboard/run/left-panel/component.js
```diff
@@ -110,38 +109,30 @@ export default Component.extend(FullScreenMixin, {
...
+            })
```
bodom0015: Tale should already be available via the `this.model.tale` (and will be `this.model` after the 0.8 refactoring). Is there a reason to requery for it here?

ThomasThelen: Since `this.model` is the instance, I'm pretty sure it only has the ID of the Tale. To make sure, I tried to log `this.model.tale` just before the request and I got `undefined`

bodom0015: Ahh, my apologies... the Run controller grabs all of this data, but it appears to be missing a line [here](https://github.com/whole-tale/dashboard/blob/master/app/controllers/run.js#L14) to actually save the Tale model that we fetch.

In the refactored Run view this should be available via the model, so I can address tidying up on that branch.

PR #478, opened 2019-05-10
Refactor Browse view to match new mockups

bodom0015:
# TODO
* Clean up dead code (list view?)
* Testing and more testing

______________________________________________________________________

### Problem
Current styling/layout differs from the mockups of the "Browse" view. 

### Approach
* Restyle cards to match mockups
* Add tab selector for future views (Data + Environments)
* Replace filter dropdown with simple buttons (currently not functional, but should be easy to wire up)
* Add helper text to the top of the view
* Remove `tale-instances` panel from the right side

### How to Test
Prerequisites: at least one Tale created, less than 2 Tales running, clear local storage (for good measure)

    * You should be brought to the new browse view, which should nearly match the [f2018 mockups](http://www.crc.nd.edu/~kfurse/wholetale-css-mock-up-sui_f2018/browse.html)
    * You should no longer see the `tale-instances` panel on the right side of the browse view
3. Choose a Tale and click `Run Tale` beneath it
    * You should see a spinner appear over the button, indicating that your Tale is launching
    * You should also see the familiar notification panel, indicating that your Tale is launching
4. Wait for the Tale to finish launching
    * You should see the spinner change to `Stop Tale` after it is done launching
5. Hover over the launched Tale's illustration and click the "View" button that appears
    * You should be taken to the Run view for the chosen Tale
6. Click "Return to Dashboard" at the top-left
    * You should be brought back to the Browse view
7. Click `Run Tale` beneath one or two more Tales (enough to trigger a launch error)
    * You should see the launch error appear in a red error label beneath the button
8. Click `Stop Tale` beneath one of your running Tales
    * A spinner should appear briefly over the button, indicating that the Tale is stopping
    * NOTE: this delay is purely artificial, as the API returns immediately from DELETE /instance calls
9. Wait a few seconds
    * The spinner should disappear after ~6 seconds, leaving you with the original `Run Tale` state


craig-willis:
Thanks @bodom0015  -- this looks great.  Things worked as expected for my basic functional testing. 

I do have a few comments/observations:
* I find the disabled "Data Catalog" and "Compute Environments" tabs and related menu items under "Create New..." to be confusing. You mentioned that this was by design. I personally think we should just hide them for v0.8 if not functional.
* I think we should include "My Tales" support for v0.8, although not as part of this PR
* The tale "title" was clickable in the original mockups, but isn't in this version.  It makes me wonder if we should make the whole upper portion of the card clickable? (@kdavis-nd) 
* For me the "Too many tales" error message overflows the card, which may be by design?

Other observations not specifically related to this PR that we can discuss or track in separate issues:
* As you noted, the "Copy" operation during copy on launch takes some time. I expect to have some indication of activity. (@Xarthisius?)
* Run view has no clear indication that I'm accessing my copy of a tale -- I can only distinguish my copy from the original on the Browse page. (@kdavis-nd?)
* Also something we've noted before, but clicking run on the original (read-only) LIGO tale does not prompt to make a copy.  (I also noticed that I could write in the workspace, which was surprising but may be local issue).

Since @kdavis-nd doesn't have access to a running instance, I've created a quick recording to show the current state of things. Unfortunately it doesn't capture the "ban circle" cursor on the tabs and menu items:

http://recordit.co/2eTqPGWBuq

kdavis-nd:
Hi - responding to the email since I’m not a member of whole-tale/dashboard. Let me know if there’s a better way to comment.

> On Jul 10, 2019, at 12:09 PM, Craig Willis <notifications@github.com> wrote:
> 
> Thanks @bodom0015 <https://github.com/bodom0015> -- this looks great. Things worked as expected for my basic functional testing.
> 
> I do have a few comments/observations:
> 
> I find the disabled "Data Catalog" and "Compute Environments" tabs and related menu items under "Create New..." to be confusing. You mentioned that this was by design. I personally think we should just hide them for v0.8 if not functional.
Can’t see the ban circles in the video, as you noted, but agreed on principle. If they can’t be hidden, maybe gray them out to indicate they’re disabled instead of adding a ban icon?

> I think we should include "My Tales" support for v0.8, although not as part of this PR
> The tale "title" was clickable in the original mockups, but isn't in this version. It makes me wonder if we should make the whole upper portion of the card clickable? (@kdavis-nd <https://github.com/kdavis-nd>)

The nice thing about the semantic ui dimmable-image-is-a-link function is that it allows you to label a button in the middle - here “View.” I think this may help clarify that both running and non-running tales can be viewed, the only difference is you can interact/edit the former and just look through the latter (right?).

Having a clickable title does not give you any of this information - by clicking this am I copying/running a tale or just viewing it? I don’t think it makes sense for the top part (with, for example, Original Tale/PUBLIC) to be clickable, and making the whole card dim with a View button seems excessive.

WWCOD - I see CodeOcean has both image and title clickable to open a capsule. Looks like they no longer automatically make a copy, but let you view the capsule and decide whether to make your own copy to edit. btw - on an unrelated note, I see they now have a nice timeline running down the right hand side of a capsule that shows its history...

I’d stick with the dimmable image link for now, and not make the title clickable, so users aren’t left to wonder if clicking the image vs the title lead to different actions. 

> For me the "Too many tales" error message overflows the card, which may be by design?
> Other observations not specifically related to this PR that we can discuss or track in separate issues:
> 
> As you noted, the "Copy" operation during copy on launch takes some time. I expect to have some indication of activity. (@Xarthisius <https://github.com/Xarthisius>?)
> Run view has no clear indication that I'm accessing my copy of a tale -- I can only distinguish my copy from the original on the Browse page. (@kdavis-nd <https://github.com/kdavis-nd>?)

Good point. This info should be incorporated into the run view, maybe somewhere around the Icon/title/name top section so it is persistent across sub-tabs. I’ll think about where it should go and what it should look like.

bodom0015:
@craig-willis regarding comments/observations:

* > I find the disabled "Data Catalog" and "Compute Environments" tabs and related menu items under "Create New..." to be confusing. You mentioned that this was by design. I personally think we should just hide them for v0.8 if not functional.

At the time, I was under the assumption that all of these views would be released together. If that is no longer the case, I should be able to hide the non-functional features to prevent user confusion.

* > I think we should include "My Tales" support for v0.8, although not as part of this PR

Noted - I am hoping that this should be relatively quick and easy to implement

* > The tale "title" was clickable in the original mockups, but isn't in this version. It makes me wonder if we should make the whole upper portion of the card clickable? (@kdavis-nd)

This was an assumption on my part that was perhaps incorrect - since this behavior was gone in the 2018 mockups and returned in the 2019 mockups, I had assumed that its absence from the 2018 mockups was an oversight. Per @kdavis-nd's recommendation, I can remove the clickable title and leave the "View" rollover intact, since that gives/could give a little more information, if that might be agreeable?

* > For me the "Too many tales" error message overflows the card, which may be by design?

Yes, this was by design to keep all of the cards a uniform height, but it can be adjusted if it is distracting or appears broken

Other things:
* > As you noted, the "Copy" operation during copy on launch takes some time. I expect to have some indication of activity. (@Xarthisius?)

Waiting to hear back from @Xarthisius on this, but if there was a notification associated with the copy operation, I could display that in the panel with the rest. I think the "Build / Create Volume / Start Container" notifications already come through for COL, so there may be some wiring necessary missing from the backend to send an additional notification about copying before the ones for launch show up. 

* > Run view has no clear indication that I'm accessing my copy of a tale -- I can only distinguish my copy from the original on the Browse page. (@kdavis-nd?)

For now, I can add a small bit of placeholder text on the `wt-paddleboard`. It should be easy enough to retroactively tweak the template to match @kdavis-nd's pending mockup

* > Also something we've noted before, but clicking run on the original (read-only) LIGO tale does not prompt to make a copy. (I also noticed that I could write in the workspace, which was surprising but may be local issue).

I could've sworn that I had added the copy-on-launch stuff into this branch, but I may have botched a merge somewhere. Will go back and investigate to make sure I haven't done something silly

PR #480, opened 2019-05-14
Fix Overflowing Datasets. Fixes #468

### LINKED ISSUE #468: [Firefox] "Select Data" Panel modal overflows
This is similar to #403, but different in that the modal seems to overflow as datasets are added. This seems to work fine in Chrome.

To Reproduce: 
1. Open Firefox
2. Launch a Tale
3. Add enough datasets to overflow the page

Screen Shot 2019-04-30 at 12 53 15 PM

Another perspective, zoomed out
Screen Shot 2019-04-30 at 12 54 59 PM

ThomasThelen:
Fix for issue #468 

The right column was missing the styling the the left column has, seen [here](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/select-data-modal/template.hbs#L21)

Testing Steps: 
See attached issue; the table should no longer overflow.

### DIFF
--- a/app/components/ui/select-data-modal/template.hbs
+++ b/app/components/ui/select-data-modal/template.hbs
@@ -76,7 +76,7 @@
                 </div>
             </div>
             <div class="seven wide column">
-                <div class="ui segment">
+                <div class="ui segment" style="overflow-y:scroll;max-height:600px;">
                     <button class="ui very basic labeled icon button" {{action 'removeSelectedData'}}>
                         <i class="angle double left icon"></i>Remove Selected</button>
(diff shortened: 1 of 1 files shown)


bodom0015:
I know that it's my fault for arbitrarily choosing `600px` in the first place, but I should note that in hindsight this makes the modal slightly too tall and causes it to no longer fit on my macbook's screen.

No need to change it as a part of this PR, but if users encounter this case often enough we may need to consider shrinking that value a bit. This may involve doing a more thorough evaluation of how many pixels the "average screen" can actually fit.

Issue #472, opened 2019-04-30
[Regression] Internal State Problems (PR #326)


ThomasThelen:
The first regression test in the [test plan](https://github.com/whole-tale/wt-design-docs/blob/master/ISSUE_TEMPLATE/test_plan.md) failed.

When performing the testing steps I ran into issues at step 6 (ran into a blank blue screen)

PR #484, opened 2019-05-29
Fix blue screen when folders are deleted

ThomasThelen:
This is a small addition to the checks we have against bad internal state as a fix for issue #472.

The cause was that I had a valid currentFolderID (ie it wasn't null or undefined), but it was referencing a deleted folder. This was causing the screen to render blue when the request would fail.

Testing steps can be found under the first test under `Regression Tests` at the bottom of the [testing docs](https://github.com/whole-tale/wt-design-docs/blob/master/ISSUE_TEMPLATE/test_plan.md)

### DIFF
--- a/app/routes/manage/index.js
+++ b/app/routes/manage/index.js
@@ -44,7 +44,6 @@ export default AuthenticateRoute.extend({
 
     var folderContents = {};
     var itemContents = Deferred();
-
     if (!folderID || folderID === "null") {
       let nav = this.get('folderNavs').getFolderNavFor('home');
       folderContents = this.get('store').query('folder', {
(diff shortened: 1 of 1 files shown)


Xarthisius:
Note: In regression test item 5. if you try to upload a file instead of reloading `/manage` it silently fails, throwing some exception in the console. Not saying it should be fixed as part of this PR, but unfortunately there're still other bugs related to the state/error handling that are lurking there.

ThomasThelen:
I was tempted to observe the `currentFolderId` in the internal state service and refresh the page when it changes, but there would of course be issues when a user is navigating around in another tab.

I think the ultimate solution is moving the state out from the local storage and encapsulating it into an object; maybe we can talk about this in the next dev call.

app/routes/manage/index.js
```diff
@@ -93,6 +91,9 @@ export default AuthenticateRoute.extend({
             limit: "0"
           }
         }
+      }).catch((err) => {
+        console.log('Failed to get folder', err);
+        return;
```
bodom0015: Very minor point, but because this happens in a Promise callback/error-handler, I don't believe that this `return` statement is necessary.

PR #493, opened 2019-06-12
Improve instance polling behavior

### LINKED ISSUE #488: Instance polling continues after logout
Identified in v0.7.0rc3

If you logout while an instance is starting, the polling continues resulting in 401 errors in the console

Test case:
* Open the dev console
* Start an instance
* While it is starting, logout
* What happens: GET instance requests continue with 401 errors
* What should happen: polling should stop

Screen Shot 2019-06-07 at 12 27 52 PM

bodom0015:
### Problem
When a user launches a Tale instance, the UI will poll for its status until it finishes. If the user refreshes their view, the poller does not reactivate. Similarly if the user logs out, the poller continues to run and spit out many errors.

Fixes #488 

### Approach
I have rewritten the current looping logic into generic `waitFor` functions in the `api-call` service - these helper methods were to be introduced in the refactored run/browse view PRs, but I have fast-tracked them here instead since they seem pertinent to the issue at hand. Note that I have not replaced all existing loops/polls with this helper as of yet, as this will likely cause merge conflicts within the other refactoring branches (where this has already been done).

* On init, check if any instance in launching (if true, poll until it finishes)
* Stop polling if an error is hit

NOTE: We are still working on building out the `notification-stream` patterns, so maybe it would be best to use that to receive async updates from Girder, rather than polling at all.

### How to Test
Prerequisites: no Tales currently running or launching

1. From deploy-dev, run `./destroy_instances.py` to ensure no running/launching instances exist
2. Checkout and run this branch locally, rebuild the dashboard
3. Open an incognito window and log into the WholeTale Platform
4. Open your developer console to the Networks tab
5. Launch a Tale from the Browse view
    * You should see the Networks tab receives a `GET /api/instance/:id` response every ~1 second
6. While the Tale is still launching, refresh the view
    * You should still see that the Networks tab receives a `GET /api/instance/:id` response every ~1 second
7. While the Tale is still launching, log out of the WholeTale Platform
    * You should see a single 401 error come through
    * The polling should stop - you should only see one error per poll that was running

### DIFF
--- a/app/components/ui/tale-browser/component.js
+++ b/app/components/ui/tale-browser/component.js
@@ -93,6 +93,17 @@ export default Component.extend({
 
     component.set('addButtonLogo', '/icons/plus-sign.png');
     component.setFilter();
+    
+    // Check if any instance is LAUNCHING
+    component.get('store').findAll('instance').then(instances => {
+        instances.forEach(instance => {
+            // If any instance is LAUNCHING, poll until it finishes
+            if (!instance.status) {
+                console.log('Instance is LAUNCHING. Starting poller...');
+                component.get('apiCall').waitForInstance(instance);
+            }
+        });
+    }).catch(err => console.error(err));
   },
   didRender() {
     $('.selectable.cards .image').dimmer({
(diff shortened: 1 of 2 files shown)


Xarthisius:
@bodom0015 could you please rebase?

Issue #523, opened 2019-07-31
Dashboard displays Job as Tale during import


craig-willis:
When testing https://github.com/whole-tale/gwvolman/pull/82 I noticed a strange behavior -- a 400 error in the console (invalid tale ID) and a Tale displayed on the Browse page with title "Import Tale".  It appears that during the import flow the compose dialog is treating a Job response as a Tale.

To repeat:
* Assumes deploy-dev with Ligo tale and `["https://dev2.dataverse.org"]` in `extra_hosts` in [wholetale plugin settings](https://girder.local.wholetale.org/#plugins/wholetale/config)
* Checkout and build dashboard `master` branch
* Open dev console
* Go to https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdev2.dataverse.org%2Fdataset.xhtml%3FpersistentId%3Ddoi%3A10.5072%2FFK2%2FNYNHAM&name=Dataverse+IRC+Metrics
* While Tale is starting, select "View" on Ligo tale then "Back to Dashboard"
* Note that a Tale is displayed with title "Import Tale" and there is a 400 error with invalid tale ID message.

<img width="1660" alt="Screen Shot 2019-07-31 at 11 13 55 AM" src="https://user-images.githubusercontent.com/4334350/62224670-0bb5d900-b385-11e9-82ba-4be13a83ce1d.png">

craig-willis:
Also noting that when testing this same flow with https://github.com/whole-tale/dashboard/pull/517, I am redirected to the "Job" tale (which doesn't exist).

Xarthisius:
I believe this was fixed by https://github.com/whole-tale/girder_wholetale/pull/347 Please reopen if it's still the case.

Issue #524, opened 2019-07-31
Handle instance cap during AiWT (import)


craig-willis:
I ran into another "too many instances" case when testing  with https://github.com/whole-tale/gwvolman/pull/82. This is related to issues reported when testing https://github.com/whole-tale/dashboard/pull/517#issuecomment-515532703, but I'm creating a separate issue since import uses a job that creates the instance and will require a different approach.

To repeat:
1. Checkout Dashboard `automatic-navigation-to-run-tabs`
1. Start 2 instances
2. Use AiWT: https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdev2.dataverse.org%2Fdataset.xhtml%3FpersistentId%3Ddoi%3A10.5072%2FFK2%2FNYNHAM&name=Dataverse+IRC+Metrics
3. Select "Create and Launch"

Expected results:
* Tale is created and too many instances error message is displayed to the user

Actual results
* Tale is created and instance is not started. Error message is not displayed to the user, so they have no idea why.

We'll need to consider options, since the import is in a Job. It may be that the only way to handle the error is through the notification stream or a pre-check prior to create and launch.

bodom0015:
Fixed by #591

Issue #522, opened 2019-07-30
Missing Instance Limit Warning when Creating Tale Copy


ThomasThelen:
The user doesn't see an error message when doing a copy + launch when there are two instances running.

To Reproduce:
1. Make sure you have two instances running
2. Launch a Tale that you don't own
3. Note that you don't receive an error about the instance limit

PR #517, opened 2019-07-23
Automatic navigation to appropriate Run tabs

bodom0015:
### Problem
Under certain conditions, the user doesn't necessarily know where they should go next.

Fixes #508 
Fixes #507 
Fixes #506 
Fixes #505 

### Approach
Help "guide" the user to where we think they want to go:
* Disable the "Run > Interact" tab when the Tale is not running
* When a user creates a new Tale from the Browse view, they are taken to the "Run > Interact" tab if the user chose to Start the Tale or "Run > Metadata" if they chose not to start it
* When a user clicks "Run" from the Browse view, they are taken to the "Run > Interact" tab for the now launching Tale
* When a user clicks "View" from the Browse view, they are taken to the "Run > Interact" tab if a Tale instance exists, or to "Run > Metadata" tab for a stopped Tale
* When a user clicks "Run" from the Run view, they are taken to the "Run > Interact" tab for the now launching Tale

### How to Test
Prerequisite: At least one stopped Tale, at least one Tale you don't own

1. Checkout and run this branch locally, rebuild the dashboard
2. Login to the WholeTale Dashboard
3. Navigate to the Browse view and choose a stopped Tale
4. Browse->Metadata 1: Create Tale" (without launching) from Create Tale modal
   * You should be brought to the "Run > Metadata" view for this Tale (#508)
   * You should see that the "Run > Interact" tab is disabled (#506)
5. Browse->Metadata 2: Hover over the Tale image and click the "View" button that appears on hover
   * You should be brought to the "Run > Metadata" view for this Tale (#505)
   * You should see that the "Run > Interact" tab is disabled (#506)
6. Browse->Interact 1: "Create Tale And Launch" from Create Tale modal
   * You should be brought to the "Run > Metadata" view for this Tale (#508)
   * You should see that the "Run > Interact" tab is disabled (#506)
7. Browse->Interact 2: Go back to the Browse view and click the "Run" button beneath a stopped Tale
    * You should be brought immediately to the "Run > Interact" view (#507)
    * You should see that the Interact tab is now enabled (#506)
    * You should see the status message(s) indicating that your Tale is now launching
8. Browse->Interact 3: Hover over the now-launching Tale's image and click the "View" button that appears on hover
    * You should be brought immediately to the "Run > Interact" view (#507)
    * You should see that the Interact tab is now enabled (#506)
    * You should see the status message(s) indicating that your Tale is now launching
9. On the "Run > Interact" view, wait for the Tale ot start and then click the "Stop" button to stop it again
[...]

PR #525, opened 2019-08-05
Use the taleFromDataset from the api-call service

ThomasThelen:
I'm not sure what the _real_ problem is for issue #523 , but I found a workaround.

The offending area seems to be the [importTakeFromDataset](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L149) method. I tested against old UI and backend versions and things seemed to be working as expected until this change (for example, the old compose route wasn't doing this).

My fix is to do it the same way that the compose page was doing it, using [taleFromDataset](https://github.com/whole-tale/dashboard/blob/master/app/services/api-call.js#L164) in the `api-call` service (Example [here](https://github.com/whole-tale/dashboard/blob/454_infinite_recursion/app/components/dashboard/compose/left-panel/component.js#L342)). This is definitely less Ember-y than what was going on in the create-tale-modal, but it doesn't have that issue where the dashboard is creating a volatile record with details from the import job.

### DIFF
--- a/app/components/ui/create-tale-modal/component.js
+++ b/app/components/ui/create-tale-modal/component.js
@@ -6,7 +6,7 @@ import $ from 'jquery';
 export default Component.extend({
   store: service(),
   router: service(),
-  
+  apiCall: service('api-call'),
   title: "",
   imageId: "",
   dataSet: null,
@@ -126,7 +126,6 @@ export default Component.extend({
     const self = this;
     newTale.save().then(item => {
       self.sendAction('_refresh', item);
-
       let taleId = item.id;
 
       if (self.createAndLaunch) {
@@ -147,25 +146,25 @@ export default Component.extend({
   // Does nothing with the job that is returned (see comment below)
   // ---------------------------------------------------------------------------------
   importTaleFromDataset(title, imageId, url, datasetAPI) {
-    let taleKwargs = {title};
+    let self = this; 
+    let taleKwargs = { title };
 
     let lookupKwargs = {};
     if (datasetAPI) {
       lookupKwargs.base_url = datasetAPI;
     }
+  let onSuccess = () => { self.closeModal(); };
+  let onFail = (e) => { self.handleError({responseJSON:{message:e+""}}); };
 
-    let appendPath = 'import';
-    let newTaleImport = this.get('store').createRecord('tale');
-
-    let queryParams = {taleKwargs:JSON.stringify(taleKwargs), lookupKwargs:JSON.stringify(lookupKwargs), imageId, url, spawn:this.createAndLaunch};
-    let adapterOptions = {appendPath, queryParams};
-    
-    const self = this;
-    newTaleImport.save({adapterOptions}).then(() => {
-      self.closeModal();
-    }).catch(e => {
-      self.handleError({responseJSON:{message:e+""}});
-    });
+  this.get('apiCall').taleFromDataset(
+    imageId,
+    url,
+    this.createAndLaunch,
+    JSON.stringify(lookupKwargs),
+    JSON.stringify(taleKwargs),
+    onSuccess,
+    onFail
+  );
     
     this.router.transitionTo({queryParams: {environment: null, name: null, uri: null, api: null}});
   }


Xarthisius:
Step 4. shows a different problem: Tale is not added to the list of objects that ember caches. As a result you get an empty view, while Tale is starting in the background, which might be confusing for users.

ala2

ThomasThelen:
They *should be taken to the run page via [this PR](https://github.com/whole-tale/dashboard/pull/517), but I think that we still need to refresh the view's models when the modal is closed. This is done when the filters are toggled, so it shouldn't be too hard. I'll put an issue in for it, and when 517 gets merged we can handle it. Unless we want it for 0.8.

bodom0015:
In attempting to test this with my other PR (#517), I am no longer getting error messages/redirection happening from the Browse page after creating and launching a Tale when the instance limit has been reached.

Once #517 has been merged up, we will likely need to retrofit that error-handling here due to the obvious merge conflict.

ThomasThelen:
@bodom0015 is that related to issues #522 and #524 ? These can probably be killed with the same stone (But if I recall correctly wasn't prioritized for 0.8).

craig-willis:
@ThomasThelen or @bodom0015 With #517 merged I'd like to review this along with those changes.  Resolving the conflict isn't clear to me -- could one of you rebase this?

bodom0015:
@craig-willis I don't think that the desired behavior has been made clear. Is there any chance we could get together and talk about exactly what needs to change here?

I vaguely recall this being brought up as part of the [Notifications discussion](https://docs.google.com/document/d/1IkvsZqxxlIBaxVLT1yb1kWtGyDuJRo8ZzT1hsqke_NE/edit#heading=h.jo67ut72go9b), where I believe this was one of our 6 or 7 use cases covered by the notification panel. While we have designed this functionality, it has not yet been implemented in the Notifications Panel, which currently only displays notifications for Start Tale and Rebuild Tale.

Is this a topic on which we need further UI/UX design? I'm not sure what exactly is our expected pattern for the somewhat typical case presented by AiWT, but this appears to be precisely the same case as "Start Tale" - we create an instance (which is immediately returned), and then poll for its status until it is complete. Unfortunately, the pattern used for "Start Tale" directly conflicts with our desire to immediately redirect the user when the modal is dismissed, as the new Tale does not yet exist. So "resolving the conflict" for this case may require us to do some design work here to further consolidate these cases into a sane pattern.

ThomasThelen:
If I recall correctly, we lumped import into a single job because we went along with the progress bar design on the compose page (which allowed us to put the progress notifications in a single job stream). 

Looking at the [job in gwvolman](https://github.com/whole-tale/gwvolman/blob/master/gwvolman/tasks.py#L503), it looks like we should be able to create the Tale before handling any registration, and then updating the Tale once complete in a potentially separate job. I agree that we should get together and chat about this before making any changes.

craig-willis:
@bodom0015 I may be misunderstanding something here, but my expectation is that in fixing #523 we wouldn't treat the job as a tale (i.e., do not add the job object to ember cache, do not display it on the browse page, do not redirect to job ID in place of tale ID).

We can address #524 later, since it is currently a problem in v0.7 and would require more thinking about notification handling.

Is there more to this than just not interpreting the job response object as a tale?

ThomasThelen:
On one level, the issue is that the response from the `tale/import` endpoint is a job so we have no way of redirecting to the run page until the [job completes](https://github.com/whole-tale/gwvolman/blob/master/gwvolman/tasks.py#L611) (because this is the final return value). This is in contrast to the `tale/` endpoint that returns a tale, which is what we get [here](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L151).
To clear up any confusion, [this](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L194) line incorrectly calls the response a Tale when it's really a job.

Ember may be getting confused because we tell it to create Tale model [here]() but actually send it to the `tale/import` endpoint via the [adapterOptions](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L191)-feeding it a job model instead of Tale.

On the backend though, the Tale is created. When we visit the browse page on a refresh, the Tales are retrieved, grabbing the newly created (correct) Tale and showing it in the UI. That's at least what I think is going on.

Bringing us back to the top, I think that we need `tale/import` to return an almost empty Tale model while a job in the background fires off that handles populating the Tale with registered data (I think this is what Mike is suggesting). This allows us to redirect to the run page with a proper Tale ID while the instance is started and registered data is added.

craig-willis:
Thanks, @ThomasThelen. I think I'm clearer now.  When we moved the Compose functionality to the Create Tale model and Browse page, we didn't bring along the import handling.  I understand @bodom0015's comment now and it's clear that we need to discuss this further as it likely does require additional changes.

ThomasThelen:
As per https://github.com/whole-tale/girder_wholetale/pull/347

we can close this since we're getting a Tale model back. Following the testing steps in 347 results in proper redirect behavior on the dashboard master branch

Issue #528, opened 2019-09-04
Run button enabled while instance is starting, tale UI not displayed without refresh


craig-willis:
I ran into this while testing https://github.com/whole-tale/girder_wholetale/pull/347, but it seems to be independent of that PR

Test case:
* Create a new tale and select "Create and Launch New Tale"
* You should be taken to the Run page/Interact tab for the new tale
* Dismiss the notification
* Note that the "Run" button is enabled throughout, even after the tale has started
* Note that the "Please Wait" spinner continues to display even after the tale has started
* Refresh
* Note that Stop button is displayed, iframed Tale UI displays in Interact tab

PR #530, opened 2019-09-16
Run button fix

ThomasThelen:
Bug fix for issue #528 

The bottom line bug was that we weren't checking for status 0 on the instance object. There was an edge case where the instance wasn't defined on the Tale object when coming from newly created Tales so some code changes were made in that area.

### To Test
Refer to #528 

1. Create and launch a Tale
2. Note the Stop button
3. Wait for the Tale to finish launching
4. Stop the Tale
5. Navigate to Browse'
6. Launch the Tale
7. Note the Stop button

### Create Tale Modal Changes

In [other places](https://github.com/whole-tale/dashboard/search?q=startTale&unscoped_q=startTale) in the codebase we use [api-call.startTale](https://github.com/whole-tale/dashboard/blob/master/app/services/api-call.js#L583) to start Tales. This handles instance creation and returns the instance. I changed this component to use `startTale` and then set the `instance` property on the Tale to the new one (this is important).

### View Route Changes

We were returning a Tale object that we construct with [api-call.fetchTaleData](https://github.com/whole-tale/dashboard/blob/master/app/services/api-call.js#L635). I'm not 100% sure why we were using this, but not using it didn't seem to break anything. This method attaches an instance to the Tale object. One issue that I ran across was that the instance wasn't yet defined on the Tale object by the time we got to the Run page. To get around this, I set the instance in the Create Tale Modal after the call to `startTale` (matching what's done in other places). Additionally I just return the actual Tale object from the Store rather than the piecewise constructed Tale we were returning before.

The `currentInstanceId` in the local storage was getting set to the Tale ID so I took this logic out and moved it to the init method in the Run page. This led to me removing the code referencing `hasSelectedTaleInstance`, which was an artifact left over when we could select instances on the right hand side of the UI. I left the `currentInstanceId` in because it gets used elsewhere in the dashboard.

### Run Page Changes

We were missing a check for instances that had a state of 0 (which is starting). I added a check for this in the template.

### What Happens When You Stop a Starting Instance?

When a user clicks the big `Stop` button, [stopTale](https://github.com/whole-tale/dashboard/blob/630290cadda30a0dee2a776e661581b64ed26bfc/app/components/dashboard/run/left-panel/component.js#L412) in the component gets called. This method ends up calling [stopTale](https://github.com/whole-tale/dashboard/blob/2feeb00ee041abc809be0941ceffb63de3681921/app/services/api-call.js#L608) in the api-call.js file. Right now we aren't doing anything when the instance status is 0 (Starting) or 2 (Failed). Is this something we want to fix in 0.8?

### Notes on Commit: "Remove references to currentInstanceId" 

While working on the view route of the run page, I moved a line of code that set `currentInstanceId` to the component, which broke things on a page refresh. This local storage variable was used in a previous version of the dashboard where the user could select a list of running Tales on the right hand side of the UI. 
The quickest fix was to just remove all references to `currentInstanceId`. This included

1. Deleting app/components/ui/tale-instances
We no longer use this component in the dashboard
2. Removing a number of commented out references
3. Changing the check in app/components/ui/tale-tabs-selector/template.hbs
I'm pretty sure this is making sure that the embedded iframe instance is the one that was selected on the right hand side list (that no longer exists).
The outer 'and' statement should now read "if there's an instance and it has an iframe"


app/routes/run/view.js
```diff
@@ -3,14 +3,15 @@ import { inject as service } from '@ember/service';
...
   internalState: service(),
-  apiCall: service('api-call'),
+  queryParams: {auth: false},
+  store: service(),
+  
   model(params) {
     this._super(...arguments);
-    this.get('internalState').set('currentInstanceId', params.tale_id);
-    //this.set('queryParams', params);
-    return this.apiCall.fetchTaleData(params.tale_id);
```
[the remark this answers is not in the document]
ThomasThelen: I'm actually not sure what the original intent was here, but the Tale model obtained from the store has all of the information that `fetchTaleData` appends.

ThomasThelen:
I have a little blurb in the PR about what happens when you click `Stop` (you get the error mentioned above). I would prefer we handle the stop logic in another PR; from the comment it looks like we need to wait until the instance is fully created, and then delete it. Right now it attempts to delete the instance, which hasn't been fully created.

app/routes/run/view.js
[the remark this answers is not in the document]
bodom0015: Interesting.... this used to all be manually done in the `controllers/run/view.js` file. On navigating to the view, the Tale would be fetched and we would need further context for display - things like attaching the creator and the instances and other associated records.

I refactored the huge block (previously known as `grabData`) into a new reusable function called `fetchTaleData` in `api-call.js`. It is possible that this logic was redundant even before I had refactored it, further emphasizing our need for a clean slate in terms of technical debt.

If things are still working as expected and displaying without error, then maybe this is no longer needed...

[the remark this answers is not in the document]
ThomasThelen: Ah that code might have pre-dated some of the attribute changes on the backend Tale model. I believe that, currently, _every_ Tale will have these attributes and there isn't a need to attach them anymore.

app/routes/run/view.js
```diff
@@ -3,14 +3,15 @@ import { inject as service } from '@ember/service';
...
+    return this.store.findRecord('tale', params.tale_id, {
+      return: true
```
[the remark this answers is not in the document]
ThomasThelen: Oops the was a mistake on my end. I'll replace it with `reload: true`; I wanted to make sure that we're retrieving the latest Tale model.

app/components/ui/create-tale-modal/component.js
```diff
@@ -156,9 +155,8 @@ export default Component.extend({
       self.closeModal();
 
       if (self.createAndLaunch) {
-        let newInstance = store.createRecord('instance');
-        return newInstance.save({adapterOptions:{queryParams:{imageId, taleId}}}).then(instance => {
-          this.newInstance = instance;
+        this.get('apiCall').startTale(this.newTale).then((instance) => {
+          newTale.set('instance', instance)
```
bodom0015: I think that manually calling `waitForInstance` is necessary after calling `startTale`, otherwise the instance polling will not start.

ThomasThelen: Is it okay if we defer that to another PR? We'll also want to do this when we import Tales in the same component. Unless I've broken it in this PR, then I'm more than happy to do it here

app/components/dashboard/run/left-panel/component.js
```diff
@@ -61,12 +61,7 @@ export default Component.extend(FullScreenMixin, {
     init() {
         this._super(...arguments);
         let state = this.get('internalState');
-        let shouldButtonsAppear = state.currentInstanceId;
-        if (shouldButtonsAppear) {
-            this.set('hasSelectedTaleInstance', true);
-        } else {
-            this.set('hasSelectedTaleInstance', false);
-        }
+        state.set('currentInstanceId', this.get('model').get('instance')._id);
```
bodom0015: I am getting an error from this line when refreshing the "Run" view.

For example - upon navigating to https://dashboard.local.wholetale.org/run/5d6072a40b7426cac492ad17, the view never renders and I get an error in the console:
```
TypeError: Cannot read property '_id' of undefined
    at o.init (wholetale-e526907ced1132f1aa2e27112b55b6ed.js:80)
    at o.n [as init] (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:3148)
    at new o (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:3057)
    at Function.u.create (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:3066)
    at e.create (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:1708)
    at n.create (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:2060)
    at Object.evaluate (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:1257)
    at e.evaluate (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:1146)
    at e.evaluateSyscall (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:1387)
    at e.evaluateInner (vendor-d660d4a9eb3aed0fe0d41d8d4c22d229.js:1381)
```

app/components/ui/create-tale-modal/component.js
bodom0015: The more I look at this, the more I can't help but think that this issue was purely a side-effect of the missing poller (#531). I am confused on how displaying the Stop button is the correct fix for this issue if we know the button won't work.

I think I now understand how I may have broken this, and I suspect that this what happened:
1. In PR #517, the poller was replaced with a call to `router.transitionTo` with the assumption that navigating to a new view would trigger a refresh and/or activate the poller, so the explicit calls to `waitForInstance` were removed - this was a obviously mistake
2. After navigation, the poller is no longer automatically refreshing instance state behind the scenes. Because we are no longer polling for the instance state (which controls the state of the "Run" button), the user's view never updates to show the started instance unless they navigate themselves (triggering a refresh)
3. Finally, we still get the event in the NotificationStream telling us that the container has started, but the component never fetches the new instance status and as a result the view never updates

As far as I can tell, activating the poller in all applicable cases will fix this issue. I am also a bit wary of removing the `fetchTaleData` stuff right now, as it is unclear if it is in fact still needed and/or if removing it would have any negative side-effects.

app/routes/run/view.js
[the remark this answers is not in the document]
bodom0015: I am still a bit wary of removing the `fetchTaleData` stuff right now, as it is unclear if it is in fact still needed and/or if removing it would have any negative side-effects.

In fact, I see one right now - there is no Creator listed under my newly-created Tale:
Screen Shot 2019-09-18 at 2 54 18 PM

Compared with the same test on dashboard.stage.wholetale.org:
Screen Shot 2019-09-18 at 2 57 08 PM

Without fetching the `creator`, we won't have anything to display beneath the Tale's title. Similarly, we need to fetch the `image` for the metadata page to properly display attributes about the Tale's environment.

app/components/ui/create-tale-modal/component.js
bodom0015: I've started testing this on another branch here: https://github.com/whole-tale/dashboard/compare/master...fix-instance-poller

I may have still missed triggering the poller in one or more cases, such as import, but will try to gather them all there. There are a lot of these cases, and they're all unfortunately (and usually unnecessarily) unique.

app/components/dashboard/run/left-panel/component.js
ThomasThelen: I think I've found a way to totally remove `currentInstanceId` from the dashboard, which should fix this.

Issue #531, opened 2019-09-18
Instance polling during Tale startup no longer functions as expected


bodom0015:
In all cases, instance polling appears to no longer function.

Steps to Recreate:
1. `Create and Launch a Tale` OR `Launch a Tale from Browse` OR `Launch a Tale from Run`
    * You should be routed automatically to the Run view
2. While the Tale is starting, open the Dev Console to the Networks tab

Expected Behavior:
You should see a request to GET `/instance/:id` every second or two.

Actual Behavior:
You won't see any GET `/instance/:id` requests, so the view never updates unless the user navigates.

bodom0015:
Digging into this, it looks like we are checking for `modelType` instead of `_modelType`: https://github.com/whole-tale/dashboard/blob/master/app/services/api-call.js#L504

It is unclear why this behavior has changed, but it seem like the poller can't start polling if it doesn't know which endpoint it needs to poll.

bodom0015:
Fixed by #534

PR #534, opened 2019-09-18
Fix instance poller

bodom0015:
### Problem
Present in the Dashboard are a set of helper methods that I refer to as the "instance poller". As the name implies, its job is to poll an instance status and automatically stop once the status matches the target value. This is helpful during instance startup, where we need to poll until the `instance.status` is equal to `1` (Running). 

Previously, the instance poller would activate during the following states:
* When creating a new Tale from the Browse view with the "Create and Launch" option
* When importing a Tale via query-string parameters from the Browse view with the "Create and Launch" option
* When starting a Tale from the Browse view
* Upon loading the Browse view, when one or more Tales are already launching
* When performing a "Copy on Launch" from the Browse view
* When starting a Tale from the Run view
* Upon loading the Run view, when the current Tale is already launching
* When performing a "Copy on Launch" from the Run view
+ _(have I missed any other use cases here?)_

After PR #517, the instance poller no longer appears to operate. See #531 

### Approach
Change all methods of launching tales (listed above) to use the new `startTale` helper method. Ensure that `self.get('apiCall').waitForInstance(instance);` is explicitly called every time `startTale` is called.

### How to Test
Prerequisites: at least on public/read-only Tale to copy (can use LIGO), stop all running Tales, keep Networks tab open in browser Dev Console

#### Create Tale
1. At the top-right, choose "Create New Tale", enter a name/environment, and confirm with the "Create and Launch" option
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
    * The instance poller should be running (check Networks tab)
2. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
3. Stop all Tales to reset test state

#### Import Tale
1. [Import a tale](https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdataverse.harvard.edu%2Fdataset.xhtml%3FpersistentId%3Ddoi%3A10.7910%2FDVN%2F3MJ7IR&name=Replication+Data+for%3A+"Agricultural+Fires+and+Health+at+Birth&asTale=True) and confirm with the "Create and Launch" option
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
    * The instance poller should be running (check Networks tab)
2. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
3. Stop all Tales to reset test state

#### Launching a Tale from Browse / On Load from Browse
1. From Browse, choose "Run Tale" beneath an existing Tale
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
2. While the Tale is still launching, **open a separate tab** to the Browse view and open the Dev Console
    * The Please Wait spinners should still be shown
    * The poller should still be running (check Networks tab)
3. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
4. Stop all Tales / close separate tab(s) to reset test state

#### Copy on Launch from Browse
1. From Browse, choose "Run Tale" beneath an existing public/read-only Tale
    * You should be presented with a confirmation modal explaining "Copy on Launch"
2. Confirm the modal
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
    * The instance poller should be running (check Networks tab)
3. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
4. Stop all Tales to reset test state

#### Launching a Tale from Run / On Load from Run
1. From Browse, hover over a Tale and click View
2. Click the "Run" button at the top-right an existing Tale
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
    * The instance poller should be running (check Networks tab)
3. While the Tale is still launching, refresh the page
    * The Please Wait spinners should still be shown
    * The poller should still be running (check Networks tab)
4. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
5. Stop all Tales to reset test state

#### Copy on Launch from Run
1. From Browse, hover over an existing public/read-only Tale and click View
1. From Run, click "Run" at the top-right of the page 
    * You should be presented with a confirmation modal explaining "Copy on Launch"
2. Click the "Run" button at the top-right an existing Tale
    * You should be automatically navigated to the "Interact" tab of the "Run" view
    * The Interact tab should display a loading spinner that reads "Please Wait"
    * The "Run" button at the top-right should be disabled and replaced with a spinner that reads `Please Wait...`
    * The instance poller should be running (check Networks tab)
3. Wait for the Tale to finish starting
    * The "Run"/"Please Wait" button at the top-right should now be enabled and read "Stop"
    * The instance poller should stop (Networks tab should be quiet)
4. Stop all Tales to reset test state

### DIFF
--- a/app/components/dashboard/run/left-panel/component.js
+++ b/app/components/dashboard/run/left-panel/component.js
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import FullScreenMixin from 'ember-cli-full-screen/mixins/full-screen';
 import config from '../../../../config/environment';
 import { scheduleOnce, later, cancel } from '@ember/runloop';
-import EmberObject, { computed } from '@ember/object';
+import EmberObject, { computed, observer } from '@ember/object';
 import { A } from '@ember/array';
 import { not, alias } from '@ember/object/computed';
 import $ from 'jquery';
@@ -50,6 +50,17 @@ export default Component.extend(FullScreenMixin, {
       'README.md',
       'metadata.xml'
     ],
+  
+    instancePoller: observer('model', 'model.instance', function() {
+      // If we see an instance that is "Launching", poll until it completes
+      const model = this.get('model')
+      if (model) {
+        const instance = model.get('instance');
+        if (instance && instance.get('status') === 0) {
+          this.apiCall.waitForInstance(model.instance);
+        }
+      }
+    }),
     
     repoDropdownClass: computed('publishStatus', function() {
         let status = this.publishStatus;
--- a/app/components/dashboard/run/left-panel/template.hbs
+++ b/app/components/dashboard/run/left-panel/template.hbs
@@ -29,9 +29,9 @@
                         <div class="ui red icon button" {{action 'stopTale' model}}>
                             <i class="fas fa-times"></i> Stop
                         </div>
-                    {{else if taleTransitioning}}
-                        <div class="ui blue icon button {{if disableStartStop 'disabled'}}" 
-                                disabled={{if disableStartStop 'true' 'false'}} 
+                    {{else if (or taleTransitioning (and model.instance (eq model.instance.status 0)))}}
+                        <div class="ui blue icon button {{if (or disableStartStop (and model.instance (eq model.instance.status 0))) 'disabled'}}" 
+                                disabled={{if (or disableStartStop (and model.instance (eq model.instance.status 0))) 'true' 'false'}} 
                                 {{action 'stopTale' model}}
                                 title="Tired of waiting? Click here to shutdown and try again.">
                             <i class="fas fa-spinner fa-pulse" style="padding-right:0;"></i> Please Wait...
--- a/app/components/ui/copy-on-launch-modal/component.js
+++ b/app/components/ui/copy-on-launch-modal/component.js
@@ -43,10 +43,10 @@ export default Component.extend({
               let eTaleCopy = EmberObject.create(taleCopy);
               self.apiCall.startTale(eTaleCopy).then((instance) => {
                 console.log(`Transitioning to /run/${eTaleCopy._id}`);
-                self.router.transitionTo('run.view', eTaleCopy._id);
                 self.apiCall.waitForInstance(instance).then((instance) => {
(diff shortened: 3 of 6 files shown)


bodom0015:
@craig-willis the second tab in that case was purely to verify both parts of the test:
1. The Browse view should continually refresh instance status if something is launching
2. The Run view should properly update once the instance finishes starting

The refreshing was simply to test the "On Load" behaviors. If we plan to support multiple tabs open in the future, then I agree that evolving the `notification-stream` is going to be the easier route there.

Unfortunately, as discussed before, the problem with the "Stop" case is that the API server returns before everything is fully removed. One of the following will be needed on the API side to correct this behavior:
1. The `DELETE /instance/:id` request could block (synchronous) until the container cleanup is done
2. We could add `InstanceState.DELETING` - note that in this case we will likely need special handling in the Browse view to hide ignore such instances, and further handling on the "Run" view to redirect back to browse if the instance state is `DELETING`

PR #545, opened 2019-09-23
Disable (+) button on Run > Files when Tale is readonly

### LINKED ISSUE #541: v0.8rc1 - (+) buttons enabled for read-only Tales in Run > Files
Steps to Reproduce:
1. Login to the WholeTale Dashboard
2. "View" (not "Run") a Tale that allows you to read, but not write
3. Navigate to Run > Files > Tale Workspaces
4. You should see the Tale's Workspace folders/files are listed here
5. Repeat the above for Run > Files > External Data

Expected Results:
* You should see that (+) has been greyed out, disallowing writes

Actual Results:
* In both cases (Tale Workspace / External Data), the (+) button is still enabled, and appears to offer the ability to add files/folders to resources that the user does not own

bodom0015:
## Problem
In the Run > Files view for read-only Tales, the (+) button is still enabled. This could lead the user to believe that they can freely add files, although the API operations themselves will fail due to permissions errors.

Fixes #541 

## Approach
Disable the (+) button when the Tale is not writable. There is logic already there for this, but it was unclear why it no longer functions.

## How to Test
Prerequisites: Register the LIGO Tale (or another read-only public Tale)

1. Checkout and run ths branch locally, rebuild the dashboard
3. Navigate to the Run > Files view for the LIGO Tale
4. On the left side, click the Home nav
    * You should see the (+) button at the top-right is **enabled**
4. On the left side, click the External Data nav
    * You should see the (+) button at the top-right is **disabled**
4. On the left side, click the Tale Workspace nav
    * You should see the (+) button at the top-right is **disabled**

### DIFF
--- a/app/components/ui/files/bread-crumbs/component.js
+++ b/app/components/ui/files/bread-crumbs/component.js
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
 import layout from './template';
 
-const taleStatus = Object.create({
+const TaleAccessLevel = Object.create({
   NONE: -1,
   READ: 0,
   WRITE: 1,
@@ -19,7 +19,7 @@ export default Component.extend({
   
   // Flag that can be used to tell if the current user has permission to edit the Tale
   canEditTale: computed('model._accessLevel', function () {
-        return this.get('model') && this.get('model') && this.get('model').get('_accessLevel') >= taleStatus.WRITE;
+        return this.get('model') && this.get('model') && this.get('model').get('_accessLevel') >= TaleAccessLevel.WRITE;
   }).readOnly(),
   cannotEditTale: computed.not('canEditTale').readOnly(),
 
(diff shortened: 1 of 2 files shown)


app/components/ui/files/bread-crumbs/component.js
```diff
@@ -1,12 +1,27 @@
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
+import { computed } from '@ember/object';
 import layout from './template';
 
+const taleStatus = Object.create({
```
Xarthisius: Can we rename that to `taleAccess` or just something other than `taleStatus`. I'm concerned it's gonna be confusing with an actual TaleStatus (`{PREPARING, READY, ERROR}`)

bodom0015: :+1: I just copy-pasted this from another place where we use the same pattern and am less concerned with the naming conventions of legacy code

That said, I went ahead updated the enum name to `TaleAccessLevel` (in this usage only)

Issue #415, opened 2019-02-20
Hide "Move To..." / "Copy to Home" in Tale Workspaces view


bodom0015:
These functions currently have outstanding problems that will not be fixed for this release.

We should hide them from the Tale Workspaces UI for now, until they can be investigated and fixed.

Once https://github.com/whole-tale/wt_home_dirs/issues/26 is addressed, we can look into re-enabling them.

PR #556, opened 2019-10-15
Disable 'Copy to Home'. See #415

Xarthisius:
As per #415 this option doesn't work yet somehow it found its way back to the UI.


Xarthisius:
@ThomasThelen that also addresses Home weirdness you [mentioned](https://github.com/whole-tale/wt-design-docs/issues/106#issuecomment-541954343)

bodom0015:
This option was added back due to confusion. The test plan still references this option, so I thought it that was removed in error. Apologies if this was not the case.

See `Workspaces 8i` under https://github.com/whole-tale/wt-design-docs/blob/master/ISSUE_TEMPLATE/test_plan.md#run

See https://github.com/whole-tale/dashboard/pull/546

PR #558, opened 2019-10-16
Adjust z-index on error message. Fixes #555

### LINKED ISSUE #555: Error message about too many instances on Browse page is obscured
Run 2 Tales. From the Browse page, click on "Run Tale" button on a third one:

error

Xarthisius:
Additionally this removes obsolete "Export as Zip" and changes wording from "Read the Docs" to "Learn More".

### DIFF
--- a/app/components/dashboard/run/left-panel/template.hbs
+++ b/app/components/dashboard/run/left-panel/template.hbs
@@ -61,13 +61,10 @@
                         <div class="tale-instance-menu">
                             <div class="ui vertical left menu transition">
                                 <a class="item" href="https://wholetale.readthedocs.io/en/stable/users_guide/run.html" target="_blank">
-                                    Read the docs <div class="ui label transparent"><i class="fas fa-book"></i></div>
+                                    Learn more <div class="ui label transparent"><i class="fas fa-book"></i></div>
                                 </a>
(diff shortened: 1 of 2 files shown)


Xarthisius:
@bodom0015 if there's a z-index value that's "above" Tale cards, but "below" everything else I can use it :)

bodom0015:
> The default z-index value of all the elements on a web page is auto, which corresponds to 0 where no z-index is assigned.

Since the `notification-stream` doesn't currently use z-index, it uses auto == 0 for the value. The log viewer modal uses `z-index: 1001`. We could probably assign `z-index: 1000` to the `notification-stream` as long as it wouldn't break anything else.

Xarthisius:
> > The default z-index value of all the elements on a web page is auto, which corresponds to 0 where no z-index is assigned.
> 
> Since the `notification-stream` doesn't currently explicitly set the `z-index`, it uses `auto` == `0` for the value. The log viewer modal uses `z-index: 1001` (along with likely most/all Semantic modals). We could probably assign `z-index: 1000` to the `notification-stream` as long as it wouldn't break anything else.

Since it affects pieces of code outside of this PR, I'm inclined to merge this and fix notification stream in a separate PR.

PR #563, opened 2019-10-22
Added "Connect Additional Accounts" view

bodom0015:
## Problem
We need a new view for the "Connect Additional Accounts" functionality.

Fixes #559  
Fixes #560
Fixes #561 

## Approach
* Added a User dropdown to the navbar.
* Added a new Settings view, which is accessible from the User dropdown.

Once the API is live, real data can be wired up. This could either be done as a part of this PR, or as a subsequent one.

### Demonstration
http://recordit.co/YNyn1S1JVj

## How to Test
Prerequisite: must run alongside https://github.com/whole-tale/girder_wholetale/pull/364

1. Checkout and run this branch locally, rebuild the dashboard
2. Login to the WholeTale Dashboard
    * You should see the navbar has a new "user" icon at the top-right
3. Click the "user" icon
    * You should see a dropdown expand
    * The dropdown should contain the user's first/last name and gravatar image
    * The dropdown should provide a link to the Settings page
    * The dropdown should provide an option to Logout
4. Choose the "Settings" nav link
    * You should be brought to the new "Settings" view
    * The view should match the mockups provided, but now uses live data: https://wholetale.org/wholetale-css-mockup/src/settings.html
    * NOTE: DataONE is present, but not yet functional
5. Click "Connect Account" to the right of Zenodo
    * You should see a modal appear allowing you to choose a resource server and enter an API key
6. Choose a resource server, enter an API key, then click Connect
    * The modal should close
    * You should see a new "Authorized on" entry appear beneath Zenodo
    * You should see a red "Disconnect" button beside the "Authorized on" line
7. Click the red Disconnect button
    * You should be greeted with a confirmation modal
8. Confirm the deletion
    * The modal should close
    * You should see the "Authorized on" line disappear

PR #575, opened 2019-12-02
Allow user to configure DataONE account from User Settings view

### LINKED ISSUE #573: Configure DataONE additional account
As noted in https://github.com/whole-tale/dashboard/pull/563, "DataONE is present, but not yet functional".  

* [x] Connect additional accounts page displays DataONE providers (e.g., DataONE staging) (done)
* [x] User selects "Connect" button and is redirected through ORCID flow. As a result, DataONE token is stored in `otherTokens`
* [x] User selects "Disconnect" and token/configured account information is removed.
* [ ] Callback to Girder with the DataONE JWT when a `preauthorized` token is detected.

bodom0015:
## Problem
Following #563, DataONE accounts are not properly `authorized` from the dashboard. The ORCID login leaves the DataONE in a `preauthorized` state, and never finishes the authorization process by POSTing the DataONE JWT back to the server.

Fixes #573 

## Approach
Fetch and POST the DataONE JWT back to the server when a DataONe provider is detected in a `preauthorized` state.

## How to Test
3. Navigate to the Settings view
4. Disconnect any DataONE account(s) present
5. Beside DataONE, click "Connect Account" to login via ORCID
    * `preauthorized` - you should be taken to the ORCID login page, then back to the settings view
6. After 1-2 seconds, the view should update without refreshing
    * `authorized` - you should still be on the settings view

### DIFF
--- a/app/templates/settings.hbs
+++ b/app/templates/settings.hbs
@@ -17,7 +17,6 @@
         <div class="ui segments">
           {{#each sortedProviders as |provider|}}
             <div class="ui segment">
-              {{!-- TODO: Disable when? --}}
               {{#if (or (eq provider.state 'authorized') (eq provider.state 'preauthorized'))}}
                 <button class="ui basic right floated blue button disabled" disabled="true">Connect Account</button>
               {{else if (or (eq provider.type 'bearer') (eq provider.type 'dataone'))}}
@@ -75,7 +74,7 @@
 <div id="connect-apikey-modal" class="ui modal apikey">
   <i class="close icon"></i>
   <div class="header">
-    {{!-- TODO: How to handle article appropriately? (e.g. "a/an ORCID Account") --}}
+    {{!-- TODO: Helper to handle article appropriately? (e.g. "a/an ORCID Account") --}}
     <img class="ui avatar image" src="data:image/png;base64,{{selectedProvider.logo}}"> Connect {{selectedProvider.fullName}} Account
   </div>
   <div class="content">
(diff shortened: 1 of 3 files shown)


bodom0015:
> In your test case, you include an explicit page refresh after disconnect.

That is correct, there is still a slight bug where the "Connect Account" button for DataONE does not properly re-enable after clicking "Disconnect".

> What are your thoughts on this? Should this be addressed in a separate issue?

I wasn't bothered by it, since this is an edge case of an edge case. I suspect it would be rare for anyone to Disconnect any account at all (with the exception of debugging), and DataONE is a case where revoking doesn't actually do much anyways. If you think it would be worth the effort, it can be addressed as a separate issue.

app/templates/settings.hbs
```diff
@@ -51,10 +51,14 @@
                       <a class="ui right floated disconnect button" {{action 'showConfirmDeleteModal' provider token}}>Disconnect</a>
                       
                       {{!-- TODO: Where do we information on who authorized the connection? --}}
-                      {{#if (and (eq provider.type 'apikey') token.resource_server)}}
+                      {{#if (and (and (eq provider.type 'apikey') token.resource_server) (eq provider.state 'authorized'))}}
```
bodom0015: There is no `provider.state` when `provider.type==='apikey'`. Need to remove this clause.

bodom0015:
@craig-willis I noticed a bug in the display of API key authorizations, so I went ahead and made some additional changes to prevent the user from needing to manually refresh. Seems that I was just trying to refresh the server tokens, when I should have also been refreshing the providers themselves. 

I should have been calling `refreshProviders` (which then calls `refreshUserTokens`), instead of calling`refreshUserTokens` directly.

PR #580, opened 2020-01-13
Add explicit user selection for asTale parameter

### LINKED ISSUE #579: Add explicit user selection for asTale parameter
With WT now [enabled on the Dataverse demo site](https://github.com/IQSS/dataverse/issues/6446) I'm returning to the discussion of how to handle the `asTale` flag during Analyze in WT.  Today, the Dataverse integration defaults to `asTale=true`, because it makes sense when demonstrating repo2docker-compatible datasets. But this isn't always the case.

Per the [10/24/2019 UI/UX meeting](https://docs.google.com/document/d/1eC8_81IJlYGcvuuZLHDh6ePhide3BsQaudqk5hExIUo/edit) we have mockups demonstrating explicit user selection of how to handle input data during Analyze in WT. 

Screen Shot 2019-12-19 at 12 13 34 PM

( https://whole-tale.github.io/wholetale-css-mockup/src/browse.html , select "Create Tale" > "Simulate Analyze in WT for tale-like")

bodom0015:
## Problem
In the "Analyze in WT" case, there is no explicit option for the user to declare that they want to import their Data Source as a Tale. The only way to toggle this flag is by using the URL query-string parameter, which is not very user-friendly.

FIxes #579 

## Approach
Per the mockups, add a set of Radio buttons to the "Analyze in WT" / "import" case on the `create-tale-modal`. The first radio (default) set `asTale` to false, while the second sets `asTale` to true.

There is also an option to disable the second radio button, although it is not clear under what circumstances this may be useful.

## How to Test
Preconditions: shut down any running instances

3. Navigate to https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdev2.dataverse.org%2Fdataset.xhtml%3FpersistentId%3Ddoi%3A10.5072%2FFK2%2FC074IQ&name=Dataverse%20IRC%20Metrics&environment=Jupyter%20Classic
    * You should see the `create-tale-modal` pop up automatically after the page loads
    * You should see that `Dataverse IRC Metrics` is automatically entered as the `name`
    * You should see that `Jupyter Classic` is automatically chosen as the `environment`
    * You should see that the Dataverse DOI/URI is listed as the `Data Source`
    * You should see the newly-offered radio buttons appear at the bottom of the modal
    * You should see that `READ ONLY` is selected by default - this is `asTale=false`
4. Open the Networks tab in your Development Console, then submit the form
    * You should see that a `POST /tale/import` was sent with `asTale=false` in the query-string
    * You should be navigated to the Run view for the now-launching Tale
5. Navigate to Run > Files > External Data
    * You should see `Dataverse IRC Metrics` listed here as an immutable external dataset
6. Now, navigate to https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdev2.dataverse.org%2Fdataset.xhtml%3FpersistentId%3Ddoi%3A10.5072%2FFK2%2FC074IQ&name=Dataverse%20IRC%20Metrics&environment=Jupyter%20Classic&asTale=true
    * You should see the `create-tale-modal` pop up automatically after the page loads
    * You should see that `Dataverse IRC Metrics` is automatically entered as the `name`
    * You should see that `Jupyter Classic` is automatically chosen as the `environment`
    * You should see that the Dataverse DOI/URI is listed as the `Data Source`
    * You should see the newly-offered radio buttons appear at the bottom of the modal
    * You should see that `READ/WRITE` is now selected by default - this is `asTale=true` as specified in the query string
7. Open the Networks tab in your Development Console, then submit the form
    * You should see that a `POST /tale/import` was sent with `asTale=true` in the query-string
    * You should be navigated to the Run view for the now-launching Tale
8. Navigate to Run > Files > Tale Workspace
    * You should see the contents of `Dataverse IRC Metrics` has been imported into your editable Tale Workspace

### DIFF
--- a/app/components/ui/create-tale-modal/component.js
+++ b/app/components/ui/create-tale-modal/component.js
@@ -98,8 +98,7 @@ export default Component.extend({
           // asTale defaults to false; it will only be set to true if query param 
           // is present and some form of the string "True" (case insensistive)
           if (asTale && asTale.toLowerCase() === "true") {
-            this.set('asTale', true);
-            $('#as-tale-true-chkbox').checkbox('check');
+            later(() => $('#as-tale-true-chkbox').checkbox('check'), 500);
           }
 
           let {official, nonOfficial} = this.computeEnvironments;
--- a/app/components/ui/create-tale-modal/component.js
+++ b/app/components/ui/create-tale-modal/component.js
@@ -31,13 +31,11 @@ export default Component.extend({
     const component = this;
     $('#as-tale-false-chkbox').checkbox({
       onChecked: function() {
-        console.log('Toggling asTale to false');
         component.set('asTale', false);
       }
     });
     $('#as-tale-true-chkbox').checkbox({
       onChecked: function() {
-        console.log('Toggling asTale to true');
         component.set('asTale', true);
       }
     });
--- a/app/components/ui/create-tale-modal/component.js
+++ b/app/components/ui/create-tale-modal/component.js
@@ -99,6 +99,7 @@ export default Component.extend({
           // is present and some form of the string "True" (case insensistive)
           if (asTale && asTale.toLowerCase() === "true") {
             this.set('asTale', true);
+            $('#as-tale-true-chkbox').checkbox('check');
           }
 
           let {official, nonOfficial} = this.computeEnvironments;
@@ -113,7 +114,7 @@ export default Component.extend({
           this.set('datasetAPI', api);
         }
       } catch(e) {
-        this.handleError({responseJSON:{message:e+""}});
+        this.handleError(e);
       }
     },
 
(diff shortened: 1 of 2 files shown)


bodom0015:
For context, the current "action" text reads `Treat as source dataset for analysis`. I am open to changing it if we can decide on acceptable an alternative.

Perhaps it would be more clear with something like `Treat as immutable dataset for analysis`?

app/components/ui/create-tale-modal/component.js
```diff
@@ -25,9 +26,26 @@ export default Component.extend({
   }),
   
   defaultErrorMessage: "There was an error while creating your Tale.",
+  
+  didRender() {
+    const component = this;
+    $('#as-tale-false-chkbox').checkbox({
+      onChecked: function() {
+        console.log('Toggling asTale to false');
```
Xarthisius: Can we drop `console.log` here and a few lines below?

bodom0015: Done - `console.log` lines have been dropped

Issue #585, opened 2020-01-22
Handling invalid environment during AiWT


craig-willis:
This is minor, but I had a mismatch in the environment name ("Jupyter Lab" v "JupyterLab") between the stage and local environments and noticed during testing that a helpful error message is swallowed by the error handler:

https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdoi.org%2F10.5281%2Fzenodo.1215988&name=weecology%2FPortalData&environment=XXX

<img width="902" alt="Screen Shot 2020-01-21 at 12 50 16 PM" src="https://user-images.githubusercontent.com/4334350/72919317-446fb300-3d15-11ea-99cd-daa4ecf7b88b.png">

The actual error in this case is "Could not find specified environment: XXX"

I also noticed that the "Create" button is disabled in this case.  

Possible solutions:
* Display underlying error message, since we know the cause
* In the event of a bad environment, let the user proceed with environment selection.

ThomasThelen:
It looks like the issue is [handleError](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L163). If it gets an error with a `responsJSON` field it will set the error message to `responseJSON.message`. If it gets a generic exception (which has e.message) it will use the default error message, "There was an error while creating your Tale".

I think this is to gobble up any nasty stack traces that the system gives and instead give a nice message to the user. In our case, [we create an exception](https://github.com/whole-tale/dashboard/blob/master/app/components/ui/create-tale-modal/component.js#L104) that looks identical to the ones we're trying to gobble up.

If we're fine with setting the error message without using `handleError`, which may be counter-style I have a branch [here](https://github.com/whole-tale/dashboard/compare/missing_env_error) that I can turn into a PR.

PR #591, opened 2020-02-10
Check for e.message in Create New Tale Modal

ThomasThelen:
Fixes https://github.com/whole-tale/dashboard/issues/585 and https://github.com/whole-tale/dashboard/issues/524

I originally had this as a fix for 585 but realized the root issue is that girder will send errors back as `e.message`. Checking for this gets us the desired message.

Note that the error message isn't removed when the user selects an environment. We don't know what kind of error we get back, so we have no way to know when to clear the error message. For example, we wouldn't want to clear an authentication error when the user selects a valid environment.

To Test:
1. Deploy locally
1. Make sure you have 2 Tales running
1. Visit https://dashboard.local.wholetale.org/browse?uri=https%3A%2F%2Fdoi.org%2F10.5281%2Fzenodo.1215988&name=weecology%2FPortalData&environment=XXX
1. Note that you see an error about the environment not found
1. Pick an environment
1. Also note that you can still interact with the modal
1. Create and launch the Tale
1. See the notification about having two instances

### DIFF
--- a/app/components/ui/create-tale-modal/component.js
+++ b/app/components/ui/create-tale-modal/component.js
@@ -98,17 +98,17 @@ export default Component.extend({
           if (asTale && asTale.toLowerCase() === "true") {
             later(() => $('#as-tale-true-chkbox').checkbox('check'), 500);
           }
-
+          this.set('datasetAPI', api);
+          
           let {official, nonOfficial} = this.computeEnvironments;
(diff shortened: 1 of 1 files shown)


craig-willis:
This looks good.  Similar to @bodom0015's comment on https://github.com/whole-tale/dashboard/pull/593#pullrequestreview-359998329, when I get too many instances I have to refresh to see the Tale.

PR #592, opened 2020-02-11
Enable import notifications

Xarthisius:
This PR adds support for generic Girder `progress` notification for `import_tale` and `import_binder` tasks. It depends on https://github.com/whole-tale/girder_wholetale/pull/389

## How to test?
1. Deploy along with https://github.com/whole-tale/girder_wholetale/pull/389
2. Follow test plan from https://github.com/whole-tale/girder_wholetale/pull/389

### DIFF
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -42,10 +42,10 @@ export default Component.extend({
         const self = this;
 
         const event = self.get('selectedEvent');
-        if (event.json.data.resource.jobs != null) {
-            var jobId = event.json.data.resource.jobs[0];
+        if (event.json.data.resource.jobs && event.json.data.resource.jobs.length) {
+            let jobId = event.json.data.resource.jobs[0];
         } else {
-            var jobId = event.json.data.resource._id;
+            let jobId = event.json.data.resource._id;
         }
         self.get('store').findRecord('job', jobId).then(job => {
             if (job && job.log) {
--- a/app/components/ui/notification-stream/component.js
+++ b/app/components/ui/notification-stream/component.js
@@ -42,7 +42,12 @@ export default Component.extend({
         const self = this;
 
         const event = self.get('selectedEvent');
-        self.get('store').findRecord('job', event.json.data.resource.jobs[0]).then(job => {
+        if (event.json.data.resource.jobs != null) {
+            var jobId = event.json.data.resource.jobs[0];
+        } else {
+            var jobId = event.json.data.resource._id;
+        }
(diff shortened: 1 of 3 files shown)


app/components/ui/notification-stream/component.js
```diff
@@ -42,7 +42,12 @@ export default Component.extend({
         const self = this;
 
         const event = self.get('selectedEvent');
-        self.get('store').findRecord('job', event.json.data.resource.jobs[0]).then(job => {
+        if (event.json.data.resource.jobs != null) {
+            var jobId = event.json.data.resource.jobs[0];
```
bodom0015:
```suggestion
            let jobId = event.json.data.resource.jobs[0];
```

We don't use `var` anymore, as is not properly block scoped. Prefer to use `let` instead, where possible.

See https://blog.usejournal.com/awesome-javascript-no-more-var-working-title-999428999994

app/components/ui/notification-stream/component.js
```diff
@@ -42,7 +42,12 @@ export default Component.extend({
...
+        } else {
+            var jobId = event.json.data.resource._id;
```
bodom0015:
```suggestion
            let jobId = event.json.data.resource._id;
```

We don't use `var` anymore, as is not properly block scoped. Prefer to use `let` instead, where possible.

See https://blog.usejournal.com/awesome-javascript-no-more-var-working-title-999428999994

app/components/ui/notification-stream/component.js
```diff
@@ -42,7 +42,12 @@ export default Component.extend({
         const self = this;
 
         const event = self.get('selectedEvent');
-        self.get('store').findRecord('job', event.json.data.resource.jobs[0]).then(job => {
+        if (event.json.data.resource.jobs != null) {
```
bodom0015:
```suggestion
        if (event.json.data.resource.jobs && event.json.data.resource.jobs.length) {
```

This would throw an error for an empty array, as `[]` is not equal to `null`.
May be better to check for a truthy length instead, as that would guarantee that there is indeed at least one item in the list

Xarthisius:
> So you can use `Create New Tale` (without launching) to save yourself a lot of time.
My kindred spirit! Let's take the revolution a one step further and make it a default action in the modal!

craig-willis:
As discussed, even after `make clean & make dev` I'm seeing odd behaviors.  I usually see two notifications, one starting as `Building image` that eventually renames to `Import Tale from external dataset` and then the second notification never completes.  Here's a recent state, with a phantom third notification.  These go away on refresh:

Screen Shot 2020-02-19 at 12 23 09 PM

PR #599, opened 2020-02-20
Remember the state of filters on Browse view

### LINKED ISSUE #564: Remember the state of "All Tales" / "My Tales" on the browse page
Whenever we switch back and forth between Browse and a different view, the former is reset to "All Tales". It would be a nicer UX if we could store the information about which tab/filter was selected and apply it to the Browse view accordingly.

bodom0015:
## Problem
The Browse view doesn't remember the state of the selected filters.

Fixes #564 

## Approach
* Save filter state in `localStorage` when it changes.
* Load from `localStorage` when view initializes.

## How to Test
3. Navigate to the Browse view
    * Filter should default to `All Tales`
4. Toggle to `My Tales` and refresh the page
    * You should see that `My Tales` is selected by default
5. Click "View" on a Tale, then click the Back button
    * You should see that `My Tales` is still selected by default

### DIFF
--- a/app/components/ui/tale-browser/component.js
+++ b/app/components/ui/tale-browser/component.js
@@ -17,7 +17,7 @@ export default Component.extend({
   taleInstanceName: "",
   filteredSet: A(),
   filters: ['All', 'Mine', 'Published', 'Recent'],
-  filter: window.localStorage.getItem('browse::filter') || 'All',
+  filter: 'All',
   numberOfModels: 0,
   pageNumber: 1,
   totalPages: 1,
@@ -92,7 +92,11 @@ export default Component.extend({
     component.set('searchView', models.tales);
     component.updateModels(component, models.tales);
     component.set('addButtonLogo', '/icons/plus-sign.png');
-    component.setFilter();
+    
+    const filter = window.localStorage.getItem('browse::filter');
+    if (filter) {
+      this.actions.selectFilter.call(component, filter);
+    }
     
     // Check if any instance is LAUNCHING
     component.get('store').findAll('instance').then(instances => {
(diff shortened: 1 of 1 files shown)


craig-willis:
I may be misunderstanding, but this doesn't appear to be working for me. I tried capturing here https://recordit.co/SirrdwyySP, but it may be be entirely clear.  

If I change the filter without refreshing, then go to view or manage and back to the dashboard, the filter resets.  If I change the filter and refresh then it works as described. It seems that the refresh is persisting the filter?

ThomasThelen:
I ran through the test steps fine, but I was expecting the filter to be set without needing the refresh 
 and resetting on the refresh.

bodom0015:
Good catch! Sorry about that - I had apparently only been testing using refresh, and not by actually navigating.

@craig-willis was correct and somehow my first commit was misbehaving because the `getItem` was only being called after a refresh (and not everytime the view loaded). I think I've corrected it now if you're willing to take another look.

@ThomasThelen I'm not sure what you mean by "resetting on the refresh", as I was thinking that this feature was meant to persist between refreshes. Perhaps I am misunderstanding?

Verified that the following now works, which I think captures the full issue and all edge cases:
1. Select "Mine" as filter, click "View", then click "Back" - "Mine" is still selected
2. Navigate to "Manage" and then click "Back" - "Mine" is still selected
3. Navigate to "Settings" and then click "Back" - "Mine" is still selected
4. Refresh the view - "Mine" is still selected
5. Select "All" as filter, click "View", then click "Back" - "All" is still selected
6. Navigate to "Manage" and then click "Back" - "All" is still selected
7. Navigate to "Settings" and then click "Back" - "All" is still selected
8. Refresh the view - "All" is still selected

ThomasThelen:
I wasn't sure what the expected behavior was-but I generally expect page refreshes to reset most UI elements (which might be an incorrect assumption). If no one else thinks so, then this PR looks good to me!

PR #600, opened 2020-02-20
Increase max length of tale title on Run view paddleboard

### LINKED ISSUE #449: Tale title truncation in Run tab
When recording the v0.7 demonstration, I noticed that we're applying the same truncation to the right and left panels of the Run tab for the Tale name/title.  Ideally, the Tale title the run page header would not be truncated or have a different limit?

bodom0015:
## Problem
Tale title is truncated too early on the Run view's paddleboard.

Fixes #449 

## Approach
Increase the maximum number of characters to `100` before title is truncated.

## How to Test
3. Create a Tale with a long name, if you haven't already
4. Navigate to the Run view for this Tale
    * You should be brought to the Run view for this Tale
    * You should see that 100 characters are allowed in the title before truncation occurs

### DIFF
--- a/app/components/dashboard/run/left-panel/template.hbs
+++ b/app/components/dashboard/run/left-panel/template.hbs
@@ -12,7 +12,9 @@
                         {{#if model.copyOfTale}}
                             <a class="ui tiny blue horizontal label">COPY</a>
                         {{/if}}
-                        {{truncate-name model.title 100}}
+                        <p class="ui computer only grid" title="{{model.title}}">{{truncate-name model.title 100}}</p>
+                        <p class="ui tablet only grid" title="{{model.title}}">{{truncate-name model.title 50}}</p>
+                        <p class="ui mobile only grid" title="{{model.title}}">{{truncate-name model.title 20}}</p>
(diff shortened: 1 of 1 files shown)


Xarthisius:
I'd like to state in advance that I understand that the number of WT users running a tiling window manager is close to 0, but can we make the number of characters somehow depend on the width of the screen? See below:

screen

bodom0015:
@Xarthisius ignoring the fact that we haven't tested much with smaller screen resolutions, we can use Semantic UI's built-in CSS classes to set a max length for each screen type...

I've arbitrarily set these widths to the following for now, and we can adjust as needed:
* `computer only` - `100`
* `tablet only` - `50`
* `mobile only` - `25`

bodom0015:
@craig-willis I just guessed on the max length for each screen size, so `100` might have been too many for this case... Maybe it should be limited to `60` instead? 

If there are better numbers to use it's easy enough to change them here. The title truncation should still dynamically change with the screen size (which can be tested by resizing the window)

I will say that I don't think it's worth it to build anything further / custom for this.

Issue #597, opened 2020-02-20
Tales not shown until refresh on error


ThomasThelen:
There are two symptoms, but I think that they should have solution so I'm putting them in one issue.

#### Issue 1
When you copy + launch a Tale, but have hit the instance cap-the Tale won't be displayed in the Tale browser until you refresh the page.

#### Issue 2
When you analyze in Whole Tale, but hit the instance cap-the Tale is created but not shown in the Tale browser

bodom0015:
One lazy (yet consistent) way to handle both of these cases would be to navigate to the newly-created Tale, even though it's not running.

This creates a slight problem where the user needs to navigate back to Stop another Tale before allowing them to start this one, but at least it would make it clear that the Tale was created and it simply couldn't be started.

ThomasThelen:
If we show the Tale, it'll still be buried with all of the public ones because the user will most likely not be filtering their own. At the same time, they would be on the page where they can do everything they need to do to get their Tale running (stop one instance and start theirs).

I like the idea of redirecting them to the run page for the purposes that you brought up. @craig-willis do you want to be a third voice of reason?

PR #611, opened 2020-03-02
Redirect to run page on errors

ThomasThelen:
Redirects the user to the Tale if there is an error launching the Tale. This behaviour was built into the create-tale-modal, but we were showing the modal before loading this behavior.

This is a fix for #597 . After navigating to the run page and back the browse, users should see their Tale

### Test Steps:
1. Checkout and deploy
1. Launch two Tales
1. Copy + Launch a Tale
1. After the error message make sure you're brought to the run page
1. Navigate to Browse
1. See your Tale
1. Create and launch a new Tale (with the two instances still running)
1. After the error message note that you're brought to the run page
1. Navigate to Browse
1. See your Tale

Side Note: I found another duplicate modal issue. After viewing the run page and going back to browse... If you copy + launch again you'll get two error modals.

### DIFF
--- a/app/components/ui/copy-on-launch-modal/component.js
+++ b/app/components/ui/copy-on-launch-modal/component.js
@@ -1,5 +1,6 @@
 import Component from '@ember/component';
 import EmberObject from '@ember/object';
+import { later } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import $ from 'jquery';
 import layout from './template';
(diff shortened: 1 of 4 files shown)


app/components/ui/copy-on-launch-modal/component.js
```diff
@@ -31,7 +32,12 @@ export default Component.extend({
               let handleLaunchError = (tale, err) => {
                 console.error('Failed to launch Tale', err);
                 self.set("errorMessage", err.message);
-                $('.ui.modal.compose-error').modal('show');
+                $('.ui.modal.compose-error').modal({
```
bodom0015:
```suggestion
                $('#copy-error-modal').modal({
```

This should fix the duplicate modal issue for the copy-on-launch error case

app/components/ui/create-tale-modal/component.js
```diff
@@ -150,14 +150,12 @@ export default Component.extend({
     const self = this;
     self.set("creatingTale", false);
     self.handleError(e);
-    $('.ui.modal.compose-error').modal('show');
     $('.ui.modal.compose-error').modal({
```
bodom0015:
```suggestion
    $('#compose-error-modal').modal({
```
This should fix the duplicate modal issue for the create-tale / AiWT error cases

